PR #14, opened 2024-01-23
- move static files to different folder

tom-webber:
- setup debug for local.py existance
- modify allowed_hosts
- update pre-runserver commands


tom-webber:
purpose of this PR was achieved using whitenoise package

PR #19, opened 2024-01-24
Add sass via npm to compile stylesheets

MatMoore:
This allows us to pass variables to the design system, and will make updating easier.

Use this to apply the design system styles globally, so that parts of the search result page can be rendered from markdown. If we want to restrict this more later, we can write more specific selectors that extend
the GOV.UK styles, like this:
https://github.com/ministryofjustice/cjs_scorecard_exploratory_analysis/blob/develop/cjs_test_app/scss/markdown-styles.scss

I've set up the package.json so we manually run `npm run sass` when we want to recompile the sass. I figure this is probably good enough for us as we won't need to change stylesheets very often.


mitchdawson1982 (approved):
Looks good...

PR #25, opened 2024-01-26
Add missing main element

MatMoore:
This element is the target for the skip link.

See https://design-system.service.gov.uk/styles/page-template/


mitchdawson1982 (approved):
LGTM

PR #29, opened 2024-01-29
Dc 2982 frontend changes

PriyaBasker23:
Here is a checklist of design-related tasks to finalise on the prototype, aligning it with the specifications outlined in
https://www.figma.com/file/zTOVhHRnWfxRJqH6Gr8cIc/Catalogue-sketches?type=design&node-id=206-2398&mode=design&t=pEjdb6k2pwuuHuiw-0

completed the following task as per the following list 
- [x]  Ensure links have the govuk-link class.
- [x] Modify filters to include the desired options when initially presented to users (e.g., domain, data product).
- [x] Integrate pagination buttons at the bottom of the page.
- [x]  Remove the horizontal rule between search results or eliminate lines from the summary list component.
- [x] Incorporate the details page (accessible by clicking on a result).
- [x]  Apply Filter
- [x] Clear Filter
- [x] Remove Filters
- [x] Back buttons
- [x] Basic details page template


static/assets/js/htmx.min.js
mitchdawson1982, line 1: We are loading htmx from CDN in ./base/head.html and adding this file into the static directory. We should load it from one place only.

home/urls.py
```diff
@@ -7,4 +7,5 @@
 urlpatterns = [
     path("", views.home_view, name="home"),
     path("search", views.search_view, name="search"),
+    path("details", views.details_view, name="details"),
```
mitchdawson1982, line 10: We should refactor the details url pattern to accept an id/pk value like this. 
`path(details/<str:id>/` . 
We can then build the individual url in the search results likes this. `"{% url 'home:details' id=result.id %}"`

home/views.py
```diff
@@ -1,25 +1,99 @@
...
     client = get_catalogue_client()
-    search_results = client.search(query=query, page=page)
 
+    # Fetch domainlist without query
+    search_results = client.search(query="", page=None)
     context = {}
+    domain_list = search_results.facets['domains']
+    context["domainlist"] = domain_list
+
+    if request.GET.getlist("domain"):
```
mitchdawson1982, line 46:
```suggestion
  domains = request.GET.getlist("domain", [])
  if domains:
      selected_domain = filter_seleted_domains(domain_list, domains)
      context['selected_domain'] = selected_domain
      request.session['selected_domain'] = selected_domain
      request.session['domains'] = domains
      filter_value = [MultiSelectFilter("domains", domains)]
```

home/views.py
```diff
@@ -1,25 +1,99 @@
...
+
+        # Populated selected domain
+        selected_domain = filter_seleted_domains(domain_list, domains)
+        context['domains'] = domains
+        context['selected_domain'] = selected_domain
+
+        # Reassign to session
+        request.session['selected_domain'] = selected_domain
+        request.session['domains'] = domains
+        if not domains:
```
mitchdawson1982, line 73: Could this check be performed further up before session assignment?

pyproject.toml
```diff
@@ -9,6 +9,7 @@ readme = "README.md"
 python = "^3.11"
 django = "^5.0.1"
 pyyaml = "^6.0.1"
+htmx = "^0.0.0"
```
mitchdawson1982, line 12: Can this be removed ?

templates/base/head.html
```diff
@@ -11,5 +11,6 @@
   <link rel="manifest" href="{% static 'assets/manifest.json' %}">
   <link rel="stylesheet" type="text/css" href="{% static 'assets/css/search.css' %}">
   <link rel="stylesheet" type="text/css" href="{% static 'assets/css/base.css' %}">
+  <script src="https://unpkg.com/htmx.org@1.7.0/dist/htmx.js"></script>
```
mitchdawson1982, line 14: Can htmx be removed ?

templates/partial/search_result.html
```diff
@@ -0,0 +1,51 @@
+{% load markdown %}
+{% for result in results %}
+
+<div class="govuk-grid-row" id="search_result">
+    <div class="govuk-grid-column-full">
+      <h3 class="govuk-heading-m govuk-!-margin-bottom-2">
+          <a href="{% url 'home:details' %}?id={{result.id}}" class="govuk-link">{{result.name}}</a>
```
mitchdawson1982, line 7: We should refactor the url generation to pass the id into the function automatically.

pyproject.toml
PriyaBasker23: yes

home/views.py
PriyaBasker23: We need to reset the session with latest information . So we cant do that

mitchdawson1982 (approved):
LGTM

PR #33, opened 2024-01-30
Update catalogue library

MatMoore:
Looks like the pre-commit hook has applied some extra formatting here.

Only change is on line 32-34


mitchdawson1982 (approved):
LGTM

Issue #35, opened 2024-01-31
Feedback on first prototype


MatMoore:
## Ask designers for advice
- [ ] ~~Unsure if worth discussing but could 'sort by' be aligned right here or moved? Unsure of govuk component advice on this but I feel it would be better aligned right.~~ REVISIT LATER
- [x] URLs in descriptions - render as links in search results but not on detail page? (e.g. Use of Force)

## Just do it
- [x] Remove CSRF token for search form
- [x]  Drop the back link on the search page
- [x]  Numbers like “10000” should have commas for readability.
- [x] Hide the placeholder “feedback” button
- [x] Header captions - “Data product” not “DATA_PRODUCT”. I might have gone with govuk-caption-m rather than -l so it’s less dominating?
- [x] “Contact for Access requirement” - lowercase “a” for “access”
- [x] Don’t use title case for table headers - “Schema details” rather than “Schema Details”

## Covered in other tickets / later
1. Pagination
2. Sorting
3. Bug: applying filters clears search and vice versa
4. Shorten the URL

PR #36, opened 2024-01-31
Frontend fixes

MatMoore:
See individual commits for details.

This checks off some of https://github.com/ministryofjustice/find-moj-data/issues/35

### DIFF
--- a/templates/base/base.html
+++ b/templates/base/base.html
@@ -13,15 +13,18 @@
 {% include "base/navigation.html" %}
 
 <div class="govuk-width-container app-width-container">
-  <!--<div class="govuk-phase-banner">
+  <div class="govuk-phase-banner">
     <p class="govuk-phase-banner__content">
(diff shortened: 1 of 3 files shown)


templates/base/base.html
```diff
@@ -13,15 +13,15 @@
...
 <div class="govuk-width-container app-width-container">
-  <div class="govuk-phase-banner">
+  <!--<div class="govuk-phase-banner">
     <p class="govuk-phase-banner__content">
       <strong class="govuk-tag govuk-phase-banner__content__tag">Alpha
       </strong>
       <span class="govuk-phase-banner__text">
         This is a new service – your <a class="govuk-link" href="#">feedback</a> will help us to improve it.
       </span>
     </p>
```
jemnery, line 23: Instead of hiding the whole phase banner, I wonder if we should keep it there so it's clearly alpha, but hide the feedback link as it doesn't do anything?

We could say - "this service is under development" for the banner text?

jemnery (approved):

PR #37, opened 2024-01-31
Improve readability

MatMoore:
- Use commas in large numbers
- Format updated dates in natural language
- Remove unwanted title case
- Reduce size of result type caption and remove all caps

### DIFF
--- a/templates/partial/search_result.html
+++ b/templates/partial/search_result.html
@@ -1,4 +1,5 @@
 {% load markdown %}
+{% load humanize %}
 {% for result in results %}
 
 <div class="govuk-grid-row" id="search_result">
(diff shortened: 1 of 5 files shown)


templates/partial/search_result.html
```diff
@@ -37,7 +38,7 @@ <h3 class="govuk-heading-m govuk-!-margin-bottom-2">
         <div class="govuk-summary-list__row">
           <dt class="govuk-summary-list__key">Last updated date</dt>
           <dd class="govuk-summary-list__value">
-            {{result.last_updated}}
+            {{result.last_updated|naturaltime}}
```
jemnery, line 41: or `{{result.last_updated}} ({{result.last_updated|naturaltime}})` ?

e.g. **"Last updated date: 25th January 2023 (one week ago)"**

So the actual date is still visible. Ignore if I've not understood how the django pipe works!

jemnery (approved):

LavMatt (approved):

PR #40, opened 2024-02-01
Added sort radio

murdo-moj:
* Added a sort radio which allows users to sort by relevance, or alphabetically on name ascending or descending
* Radios are preferred over dropdowns for few options according to gov.uk
* Sort options on graphql are limited to name and relevance, urn, and time created. We might want to add time created as an option at some point if users ask for it.
* Sorting doesn't work with the pagination yet


home/views.py
```diff
@@ -29,19 +31,27 @@ def details_view(request, id):
 
 
 def search_view(request, page: str = "1"):
-    query = request.GET.get("query", "")
+
+    new_search = request.GET.get("new", "")
```
MatMoore, line 35: This is a quick fix for search queries persisting when they shouldn't.

It's bypassing behaviour caused by `request.session["query"] = query` on line 94

 I think we will be able to remove this again after some refactoring, but don't want to do this without tests!

home/views.py
```diff
@@ -78,8 +88,11 @@ def search_view(request, page: str = "1"):
             filter_value = [MultiSelectFilter("domains", domains)]
 
     elif request.GET.get("query"):
+        query = request.GET.get("query")
+        print("query", query)
```
MatMoore, line 92: Stray print here that can be removed

MatMoore (approved):

MatMoore (approved):

PR #42, opened 2024-02-02
Add query_string tag from Django 5.1

MatMoore:
This tag will be released in the next version of Django:

https://github.com/django/django/commit/e67d3580edbee1a4b58d40875293714ac3fc6937#diff-30b3434053ec4af0e27c503a53ec0e3aa8fb74ea75b458ac62507c66f2f78dd4R1172

However, it's really useful for us right now since it allows you to generate pagination links that preserve existing query parameters.

This means the search query and filters won't be lost when navigating between pages.

Any time we generate a link that goes back to the search page, we can use this to preserve the query string.

### DIFF
--- a/home/templatetags/future.py
+++ b/home/templatetags/future.py
@@ -1,0 +1,38 @@
+from django import template
+from django.utils.itercompat import is_iterable
+
+register = template.Library()
+
+
+# Backport from unreleased Django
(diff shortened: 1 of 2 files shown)


MatMoore:
Bringing this in as part of the refactor ticket

PR #62, opened 2024-02-13
Dp-3182-views-and-forms-tests

murdo-moj:
* Added django pytest support
* Added pytest tests for views
* Added pytest tests for forms
* Added a conftest which creates a mock django client and mocks the datahub catalogue
* Added a workflow to run the tests
* Added code coverage report for pytest


.github/workflows/test.yml
```diff
@@ -0,0 +1,37 @@
...
+        uses: actions/cache@v2
+        with:
+          path: .venv
+          key: pydeps-${{ hashFiles('**/poetry.lock') }}
+      - run: poetry install --no-interaction --no-root
+        if: steps.cache-deps.outputs.cache-hit != 'true'
+      - run: poetry install --no-interaction
+      - run: poetry run pytest
+      - name: coverage
+        run: poetry run pytest --cov
```
MatMoore, line 37: looks like this is running the tests twice, so you can remove the above step

core/settings.py
```diff
@@ -64,6 +64,12 @@
 
 # Database
 # https://docs.djangoproject.com/en/5.0/ref/settings/#databases
+DATABASES = {
```
PriyaBasker23, line 67: Why do we need this ?

.github/workflows/test.yml
murdo-moj: one's to generate a code coverage report; I thought it might be useful to have that as an extra step with the first test call to be more friendly to debugging. but maybe one call is enough. Ill remove the first one

home/views.py
```diff
@@ -31,8 +31,7 @@ def search_view(request, page: str = "1"):
         # Populated search scenario
         form = SearchForm(request.GET)
         if not form.is_valid():
-            print("form error on validation")
-            print(form.errors)
+            return HttpResponseBadRequest(form.errors)
```
MatMoore, line 34: I don't think this is correct - if there are form errors we should still render the page, just with an error message at the top

the design system has this page: https://design-system.service.gov.uk/patterns/validation/

I think we just skipped it for now because we didn't implement any validation yet

tests/test_forms.py
```diff
@@ -0,0 +1,44 @@
+from django.test import TestCase
+from home.forms.search import SearchForm
+
+
+class SearchFormTest(TestCase):
```
MatMoore: Separating this from the view has really helped with the tests 👍🏻

home/views.py
murdo-moj: yeah, perhaps this bit is pointless. We need to do more error handling on the form before we can return any sensible content for the error page. However it's ok as a holding pattern, and the tests weren't working for a bad form unless I returned something here (it'd just hang)

mitchdawson1982 (approved):
LGTM

PR #70, opened 2024-02-14
Add selenium tests

MatMoore:
This adds some selenium tests, using pytest-django's [live_server](https://pytest-django.readthedocs.io/en/latest/helpers.html#live-server) fixture (equivalent to Django's LiveServerTestCase)

Selenium can be really fiddly, and I don't want these tests to slow us down, so would welcome any suggestions on how to make them easy to maintain. 

(Note: the commit history is messy so this is best reviewed all together)

### Mocking
This is using the global mock @murdo-moj added for the catalogue, so the tests don't depend on datahub to run.

### Structure
I've applied the [page object pattern](https://www.selenium.dev/documentation/test_practices/encouraged/page_object_models/) so that all the selectors are in one place, so if we change the template we can change that without rewriting the whole test class.

Each step in a test extracted to a helper method so we can reuse them for multiple scenarios.

### Utilities
I've included some utilities to make this more useful:

- Run [axe-core](https://github.com/dequelabs/axe-core) against each page, to catch certain classes of accessibility issues - we had a couple of these due to mangling the html while refactoring
- Each time the driver clicks on something, save a screenshot to the tmp directory, to aid debugging

### CI
Selenium tests are quite slow, so I've made it a separate step that runs only if the unit tests pass. It uses a headless chrome browser, which is installed by default on the ubuntu runner.

## Fixes
In the process of adding these tests, I've fixed a couple of html issues such as missing labels, multiple main sections, elements appearing outside of a [landmark region](https://dequeuniversity.com/rules/axe/4.8/region).


tests/selenium/conftest.py
```diff
@@ -0,0 +1,154 @@
...
+    def screenshot(name="screenshot"):
+        nonlocal counter
+        body = selenium.find_element(By.TAG_NAME, "body")
+        body.screenshot(str(test_dir / f"{counter}-{name}.png"))
+        counter += 1
+
+    yield screenshot
+
+
+class DetailsPage:
```
murdo-moj, line 40: might be good to have a generic page base class considering all 3 pages have the same init

murdo-moj:
might we want to disable screenshots from the workflow job? we can't see any of the screenshots on the runner

tests/selenium/conftest.py
```diff
@@ -0,0 +1,156 @@
...
+
+    def __getattr__(self, name):
+        return getattr(self.element, name)
+
+    def link(self):
+        return self.element.find_element(By.CSS_SELECTOR, "h3 a")
+
+
+class SearchPage(Page):
+    def __init__(self, selenium):
```
murdo-moj, line 70:
```suggestion
```

tests/selenium/conftest.py
```diff
@@ -0,0 +1,156 @@
...
+        self.selenium = selenium
```
murdo-moj, line 71:
```suggestion
```

core/settings.py
```diff
@@ -125,3 +125,11 @@
 
 # session
 SESSION_ENGINE = "django.contrib.sessions.backends.signed_cookies"
+
+# Not actually used - Just required for LiveServerTestCase
+DATABASES = {
```
PriyaBasker23, line 130: During deployment, will this necessitate the creation of an SQLite3 database? Had  similar problem when i used sessions & database

tests/selenium/test_search_scenarios.py
```diff
@@ -0,0 +1,260 @@
...
+    @pytest.fixture(autouse=True)
+    def setup(
+        self, live_server, selenium, home_page, search_page, details_page, screenshotter
+    ):
+        self.selenium = selenium
+        self.live_server_url = live_server.url
+        self.home_page = home_page
+        self.search_page = search_page
+        self.details_page = details_page
+        self.screenshotter = screenshotter
```
PriyaBasker23, line 24: Suggestion - Can we screenshot just for test failures alone

core/settings.py
MatMoore: Nope, this is in-memory

tests/selenium/conftest.py
```diff
@@ -0,0 +1,154 @@
...
+
+
+class DetailsPage(Page):
+    def secondary_heading(self):
+        return self.selenium.find_element(By.TAG_NAME, "h2")
+
+
+class HomePage(Page):
+    def __init__(self, selenium):
```
murdo-moj, line 51:
```suggestion
```

tests/selenium/conftest.py
```diff
@@ -0,0 +1,154 @@
...
+        self.selenium = selenium
```
murdo-moj, line 52:
```suggestion
```

murdo-moj (approved):

PR #73, opened 2024-02-15
Added label for search input

murdo-moj:
* Added label for search input
* Searching the HTML for other inputs, all inputs now have a label


murdo-moj:
I gave it a go over with the screenreader and added a label for the search button.

MatMoore (approved):

PR #77, opened 2024-02-19
Include subdomains in the filter when we filter by a parent domain

MatMoore:
I've just hardcoded the domains for now, since I just want to get the top-level domain filter working as expected again after assigning assets to subdomains.


mitchdawson1982 (approved):
LGTM

PR #78, opened 2024-02-19
Find-moj-data-45/search-highlighting

murdo-moj:
* Add a matched fields entry to search results with readable names
* Apply bold highlighting for search terms on the description and search summary fields
* The search summary field isn't actually searched by datahub, it's more of a patch until we figure out how we want to handle long descriptions
* Should URN/ID even be included in matched fields? Users will not care about it's URN or ID most likely.

### DIFF
--- a/home/service/search.py
+++ b/home/service/search.py
@@ -93,12 +93,6 @@ class SearchService(GenericService):
         else:
             pattern = f'({re.escape(query)})'
             for result in highlighted_results.page_results:
-                metadata = result.metadata
-                if metadata.get("search_summary"):
-                    metadata["search_summary"] = re.sub(
-                        pattern, r'**\1**', metadata["search_summary"], flags=re.IGNORECASE,
-                    )
-
                 result.description = re.sub(
                     pattern, r'**\1**', result.description, flags=re.IGNORECASE,
                 )
--- a/home/service/search.py
+++ b/home/service/search.py
@@ -107,6 +107,8 @@ class SearchService(GenericService):
 
     def _get_match_reason_display_names(self):
         return {
+            "id": "ID",
+            "urn": "URN",
             "domains": "Domain",
             "name": "Name",
             "description": "Description",
--- a/home/service/search.py
+++ b/home/service/search.py
@@ -107,8 +107,6 @@ class SearchService(GenericService):
 
     def _get_match_reason_display_names(self):
         return {
-            "urn": "URN",
-            "id": "ID",
             "domain": "Domain",
             "name": "Name",
             "description": "Description",
--- a/home/service/search.py
+++ b/home/service/search.py
@@ -1,5 +1,5 @@
 from typing import Any
-
+from copy import deepcopy
 from data_platform_catalogue.search_types import MultiSelectFilter, SortOption
 from django.core.paginator import Paginator
 
@@ -14,6 +14,7 @@ class SearchService(GenericService):
         self.page = page
         self.client = self._get_catalogue_client()
         self.results = self._get_search_results(page, items_per_page)
+        self.highlighted_results = self._highlight_results()
         self.paginator = self._get_paginator(items_per_page)
         self.context = self._get_context()
 
@@ -42,13 +43,8 @@ class SearchService(GenericService):
             sort=sort_option,
             count=items_per_page,
         )
-        highlighted_results = (
-            self._highlight_results(results, query)
-            if query != ""
-            else results
-        )
 
-        return highlighted_results
+        return results
 
     def _get_paginator(self, items_per_page: int) -> Paginator:
         pages_list = list(range(self.results.total_results))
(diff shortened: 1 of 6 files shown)


templates/details.html
```diff
@@ -28,16 +28,16 @@ <h2 class="govuk-heading-l">{{result.name}}</h2>
...
               </div>
-      
-      
+
+
               <div class="govuk-summary-list__row">
                 <dt class="govuk-summary-list__key">Domain name</dt>
                 <dd class="govuk-summary-list__value">
-                {{result.metadata.domain.properties.name}}
+                {{result.metadata.domain.properties.name|markdown}}
```
murdo-moj, line 36:
```suggestion
                {{result.metadata.domain.properties.name}}
```

MatMoore:
I agree we can ignore matching on URN and ID

home/service/search.py
```diff
@@ -76,6 +76,42 @@ def _get_context(self) -> dict[str, Any]:
...
         }
 
         return context
+
+    def _highlight_results(self):
+        "Take a SearchResponse and add bold markdown where the query appears"
+        string_to_highlight = self.form.cleaned_data.get("query") if self.form.is_valid() else ""
+        highlighted_results = deepcopy(self.results)
+
+        if string_to_highlight == "":
```
tom-webber, line 89: Will the `self.form.cleaned_data.get("query")` return `"*"` if there's no query (as in native DataHub)? If so, should we handle it here?
```suggestion
        if string_to_highlight in ("", "*"):
```

home/service/search.py
```diff
@@ -76,6 +76,42 @@ def _get_context(self) -> dict[str, Any]:
...
+        else:
+            for result in highlighted_results.page_results:
+                metadata = getattr(result, "metadata")
+                metadata["search_summary"] = metadata.get("search_summary", "").replace(
+                    string_to_highlight, f"**{string_to_highlight}**"
+                )
+                setattr(result, "metadata", metadata)
+
+                name = getattr(result, "description")
+                highlighted_description = name.replace(
```
MatMoore, line 101: There might be conflicts here between markup added by whoever wrote the descriptions and the markup we're adding for the match highlighting.

To avoid this we could
- make sure we strip existing markdown from the description before we repurpose it for the search summaries (this is the case for the ones we've done manually
- make sure `string_to_highlight` doesn't contain any markdown significant characters (such as dashes) before using it here

MatMoore, line 101: Using string.replace here means the highlighting is going to be case sensitive. We could improve on this by using a case insensitive regex replace:

```
>>> query='bar'
>>> pattern = f'({re.escape(query)})'
>>> re.sub(pattern, r'**\1**', 'foo bar baz BAR', flags=re.IGNORECASE)
'foo **bar** baz **BAR**'
```

home/service/search.py
```diff
@@ -76,6 +76,42 @@ def _get_context(self) -> dict[str, Any]:
...
+                    string_to_highlight, f"**{string_to_highlight}**"
+                )
+                setattr(result, "description", highlighted_description)
+
+            return highlighted_results
+
+    def _get_match_reason_display_names(self):
+        return {
+            "urn": "URN",
```
MatMoore, line 110: agree we can drop urn and ID - I don't think it's going to mean much to users

home/service/search.py
murdo-moj: The "*" is set as the default query string if nothing is passed to the catalogue API library, so it'll still be a blank string in the form side afaik?

https://github.com/ministryofjustice/data-platform/blob/2622d7cbde6a34d34dfaf993088795b616f6e5da/python-libraries/data-platform-catalogue/data_platform_catalogue/client/datahub/search.py#L45

MatMoore: I think we can use either interchangably to search for everything

murdo-moj: I'm just gonna add it, I don't think there's any harm

murdo-moj:
I added back ID and URN, as otherwise some search results had blank match reasons. I've also displayed a clipped search description instead of a search summary on the search page. Pre-existing markdown will likely cause some weird rendering on the descriptions, but I couldn't think of a nice way around it other than restricting inputs for the descriptions.

I've tested the highlighted descriptions with a screenreader; it doesn't pick up the bolding.

templates/partial/search_result.html
```diff
@@ -18,10 +18,16 @@ <h3 class="govuk-heading-m govuk-!-margin-bottom-2">
           {% endif %}
       </h3>
       {% if result.description %}
-      <div class="govuk-body-m">{{result.metadata.search_summary|markdown}}</div>
+      <div class="govuk-body-m">
+        {% if result.description|length > 200 %}
+          {{ result.description|slice:":200"|add:"..."|markdown }}
```
MatMoore, line 23: A character-based slice might cause problems here depending on where it ends up slicing.  Looks like `truncatewords` might work better for this:

https://docs.djangoproject.com/en/5.0/ref/templates/builtins/#truncatewords

MatMoore (approved):

templates/partial/search_result.html
murdo-moj, line 23:
```suggestion
          {{ result.description|truncatewords:80|add:"..."|markdown }}
```

Issue #87, opened 2024-02-20
Frontend for subdomain widget


MatMoore:
We're stealing the design for this from GOV.UK. We will need to implement the behaviour in javascript. This code might be useful: https://github.com/alphagov/finder-frontend/blob/main/app/assets/javascripts/taxonomy-select.js

![Image](https://github.com/ministryofjustice/find-moj-data/assets/87579/6660c4da-f677-4dbf-a2cf-481c7981dd5b)

Figma: https://www.figma.com/file/zTOVhHRnWfxRJqH6Gr8cIc/Catalogue-sketches?type=design&node-id=203-2121&mode=design&t=WdxT6QCBFCadn78N-0

## Expected behaviour

- If javascript is disabled, do not display the subdomain widget
- It starts out in a disabled state, showing "All subdomains"
- When a domain is selected, it becomes enabled
- When "all domains" is selected for the domain, it becomes disabled again
- When a domain is selected or changed the subdomain widget is populated with the subdomains for that domain
- When subdomain is "all domains" the selected filters should show the parent domain name
- When a subdomain is selected, the selected filters should show the subdomain name, and not the parent domain (or combine into one item)
- When a subdomain filter is cleared, it resets the domain and subdomain widgets

## Out of scope
- Auto refreshing results when a subdomain is selected (we don't do this with the domain alone, and we have an explicit "apply filters" button)

![Image](https://github.com/ministryofjustice/find-moj-data/assets/87579/6a3dd8b0-5277-41d6-a80c-06132e4c39f2)

MatMoore:
https://github.com/alphagov/finder-frontend/blob/main/spec/javascripts/taxonomy_select_spec.js#L1 - tests for the GOV.UK component

https://github.com/alphagov/finder-frontend/blob/main/app/assets/javascripts/live_search.js#L169 - usage in the main search js

PR #89, opened 2024-02-20
Add javascript for domain/subdomain widget behaviour

MatMoore:
Issue: https://github.com/ministryofjustice/find-moj-data/issues/87

This is based on a similar topic/subtopic widget from GOV.UK: https://github.com/alphagov/finder-frontend/blob/ebee8ca91224eea3876490da1775aa568690f94a/app/assets/javascripts/taxonomy-select.js

This PR adds the javascript to support this widget, but does not yet wire it up to django. I will do this once @mitchdawson1982's changes are merged.

This is what it looks like with example HTML:

https://github.com/ministryofjustice/find-moj-data/assets/87579/d449094f-7443-4102-8f88-31790e835823

## Progressive enhancement
This component uses [progressive enhancement](https://www.gov.uk/service-manual/technology/using-progressive-enhancement) so that we gracefully fallback to the basic top-level domain select if the javascript fails to load for any reason.

## Testing
I've introduced [jest](https://jestjs.io/) as a javascript testing framework and integrated that with the CI. You can run it with `npm test`.

## Possible extensions
We may want to add functionality later to make other parts of the page (such as the list of selected filters) live refresh when the domain/subdomain changes.

I've not implemented this now, because we already have an "Apply filters" button and I didn't want to diverge too much from the [MOJ filter](https://design-patterns.service.justice.gov.uk/components/filter/) component before we've put this in front of users. But I've implemented a custom event so the state of this component is observable.


tests/javascript/test_enhanced_search.js
tom-webber, line 1: Could you add a test case for a domain without any subdomains?

.github/workflows/test.yml
```diff
@@ -40,6 +40,19 @@ jobs:
...
+          echo "Chromium version: $CHROMIUM_VERSION"
+          echo "CHROMIUM_VERSION=$CHROMIUM_VERSION" >> $GITHUB_ENV
+      - name: Setup Chrome 🌐
+        id: setup-chrome
+        uses: browser-actions/setup-chrome@v1
+        with:
+          chrome-version: ${{ env.CHROMIUM_VERSION }}
+      - name: Install chromedriver 🚗
+        run: |
+          npm install -g chromedriver@$CHROMIUM_VERSION
```
tom-webber, line 55:
```suggestion
          npm install -g chromedriver@${CHROMIUM_VERSION}
```
personal preference: I like to make the `env` calls stand out a little more with the curly braces

tom-webber (dismissed):
Looks great!

tom-webber (approved):

PR #91, opened 2024-02-20
Dp 3047 present additional metadata items

mitchdawson1982:
resolves ministryofjustice/data-platform#3047 & ministryofjustice/data-platform#2961

Adds
-  `sensitivityLevel` & `whereToAccessDataset` as additional filters in the UI and now passes the values of these fields into the filter for searching.
- Selected filters are moved into their own partial template and there is now logic to build them based on the clear-filter_href dictionary keys that is passed in the context.

Changes
- domain to a single select element from multi-select check box
- Clear filter link is now only displayed if there are filters to clear rather than on all requests
- `encode_without_filter` now checks for the type of value and responds appropriately

Also:
- update pre-commit
  - add sync_with_poetry hook to keep pre-commit in sync with pyproject.toml
  - config flake8
  - config isort to use black via pyproject.toml
- Add .secrets.baseline for detect-secrets
- move testing libraries to dev.dependencies in pyproject.toml


home/service/search.py
```diff
@@ -62,20 +75,29 @@ def _get_paginator(self, items_per_page: int) -> Paginator:
...
+            label_clear_href = {}
+            if domain:
+                label_clear_href["domain"] = {
+                    domain.split(":")[-1]: (self.form.encode_without_filter(domain))
+                }
+            if classifications:
+                classifications_clear_href = {}
+                for classification in classifications:
+                    classifications_clear_href[
+                        classification.split("=")[1]
```
MatMoore, line 95: It looks like the browser is sending the value back as sensitivityLevel=$value but I think if we make it use the value directly, then we wouldn't need to split it here?

If we can make this code work with arbitrary filters, then it will be easier to change which custom properties we use.

home/templatetags/clear_filter.py
```diff
@@ -0,0 +1,15 @@
+from django import template
+
+register = template.Library()
+
+
+@register.filter
```
MatMoore, line 6: I hate that django makes this necessary, but this will be useful!

templates/partial/filter.html
```diff
@@ -18,18 +19,29 @@ <h2 class="govuk-heading-m">Selected filters</h2>
...
               </p>
             </div>
           </div>
-          {% if form.domains.value %}
+          {% if label_clear_href|get_item:"domain"|length > 0 %}
             <h3 class="govuk-heading-s govuk-!-margin-bottom-0">Domain</h3>
             <ul class="moj-filter-tags">
               <input type="hidden" name="clear_label" value="">
-              {% for label, href in label_clear_href.items %}
+              {% for label, href in label_clear_href|get_items:"domain" %}
```
MatMoore, line 33: Suggestion: maybe we can move this into a partial and pass in the filter type "domain" or "classifications"

templates/partial/filter.html
```diff
@@ -18,18 +19,29 @@ <h2 class="govuk-heading-m">Selected filters</h2>
             <div class="moj-filter__heading-action">
               <p>
                 <a class="govuk-link govuk-link--no-visited-state"
-                href="{% url 'home:search' %}{% query_string domains=None %}"
+                href="{% url 'home:search' %}{% query_string domain=None classifications=None where_to_access=None %}"
```
MatMoore, line 22: Suggestion: Could we hide selected filters entirely if there are no selected filters? (Not sure if the component is designed to be used this way, but noticed it looks odd having a clear filters button with no filters)

home/forms/search.py
```diff
@@ -4,12 +4,13 @@
...
 
 
-def get_domain_choices():
+def get_domain_choices() -> list[tuple[str, str]]:
     """Make API call to obtain domain choices"""
     # TODO: pull in the domains from the catalogue client
     # facets = client.search_facets()
     # domain_list = facets.options("domains")
     return [
+        ("", "Select ..."),
```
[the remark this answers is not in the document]
mitchdawson1982: Updated.

home/forms/search.py
```diff
@@ -103,13 +129,49 @@ def __init__(self, *args, **kwargs):
...
     def encode_without_filter(self, filter_to_remove):
         """Preformat hrefs to drop individual filters"""
         # Deepcopy the cleaned data dict to avoid modifying it inplace
         query_params = deepcopy(self.cleaned_data)
-
-        query_params["domains"].remove(filter_to_remove)
-        if len(query_params["domains"]) == 0:
-            query_params.pop("domains")
-
+        domain = query_params.get("domain")
```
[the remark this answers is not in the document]
mitchdawson1982: Made an attempt at this.

home/forms/search.py
```diff
@@ -103,13 +129,49 @@ def __init__(self, *args, **kwargs):
...
         self.initial["sort"] = "relevance"
 
+    def _clean_custom_property(
+        self, custom_property: str, property_name: str
+    ) -> list[str]:
+        """Helper function to style DataHub customProperties into their search syntax
+        (customPropertyName="custom property"). Reformats the 'cleaned_data' list for a
+        property.
+
+        ref: https://datahubproject.io/docs/how/search/#advanced-queries
```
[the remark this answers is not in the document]
mitchdawson1982: No working with filters rather than as a query.

home/service/search.py
```diff
@@ -90,3 +112,35 @@ def _get_context(self) -> dict[str, Any]:
...
+            logic for custom properties
+        """
+        # ref: https://datahubproject.io/docs/how/search/#advanced-queries
+        custom_property_query: str = "/q customProperties: "
+        custom_property_strings: list[str] = []
+
+        for _, value in custom_properties.items():
+            if value:
+                # within-filter options are inclusive OR
+                custom_property_string = " OR ".join(value)
```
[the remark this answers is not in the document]
mitchdawson1982: Removed as no longer required.

templates/partial/filter.html
mitchdawson1982: Moved to partial and no longer repeats the same code for each field.

templates/partial/filter.html
mitchdawson1982: Selected filters is now hidden when there are no filters present.

home/service/search.py
```diff
@@ -62,20 +75,29 @@ def _get_paginator(self, items_per_page: int) -> Paginator:
...
+                classifications_clear_href = {}
+                for classification in classifications:
+                    classifications_clear_href[
+                        classification.split("=")[1]
+                    ] = self.form.encode_without_filter(classification)
+                label_clear_href["classifications"] = classifications_clear_href
         else:
             label_clear_href = None
-
+        print(f"label clear ref: {label_clear_href}")
```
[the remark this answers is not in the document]
mitchdawson1982: removed

home/service/search.py
```diff
@@ -31,15 +29,40 @@ def __init__(self, form: SearchForm, page: str, items_per_page: int = 20):
...
         self.context = self._get_context()
 
+    @staticmethod
+    def _build_filter_strings(
+        filter_param: str, filter_value_list: list[str]
+    ) -> list[str]:
+        return (
+            [f"{filter_param}{filter_value}" for filter_value in filter_value_list]
+            if filter_value_list
+            else []
```
MatMoore, line 40: [Minor] I don't think the conditional is needed here - if `filter_value_list` is always a list then the list comprehension will return empty if it is empty

home/service/search.py
```diff
@@ -31,15 +29,40 @@ def __init__(self, form: SearchForm, page: str, items_per_page: int = 20):
         self.paginator = self._get_paginator(items_per_page)
         self.context = self._get_context()
 
+    @staticmethod
+    def _build_filter_strings(
```
MatMoore, line 34: Could we rename this something like `_build_custom_property_filter`, and perhaps include a comment about how custom properties are indexed by datahub?

I was really confused why we were formatting like this until I remembered it was just for custom properties, so think this this is quite likely to confuse anyone reading the code.

MatMoore (dismissed):
Looks good to me 👍🏻 

Seems like there is still some complexity around building the filters and corresponding clear links but don't see any way to avoid this, and we have good test coverage now.

Once this is merged my plan is to tweak the domain filter to produce a list of filters based on both `domain` and `subdomain` params.

home/forms/search.py
```diff
@@ -4,12 +4,13 @@
 from django import forms
 
 
-def get_domain_choices():
+def get_domain_choices() -> list[tuple[str, str]]:
     """Make API call to obtain domain choices"""
     # TODO: pull in the domains from the catalogue client
     # facets = client.search_facets()
-    # domain_list = facets.options("domains")
+    # domain_list = facets.options("domain")
```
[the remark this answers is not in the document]
mitchdawson1982: done

home/service/search.py
mitchdawson1982: Done

home/service/search.py
mitchdawson1982: Done

LavMatt (approved):

PR #92, opened 2024-02-21
Dp 3072 data product detail page

LavMatt:
For: https://github.com/ministryofjustice/data-platform/issues/3072
Adds the data product details page to find-moj-data.

Includes:
- New specific html template for data product details page
- An extra arg in the href of a search link to details to establish result type and redirect to correct details service in views
- A `DetailsDataProductService` to create data product details page context
- new `ministryofjustice-data-platform-catalogue` version for added `list_data_product_assets()` search method and some other bug fixes.
- Tests for added functionality

This does not add pagination to the details of a data product, although this may want to e considered for another issue. e.g. when a data product has a large amount of associated assets


templates/details_data_product.html
```diff
@@ -91,21 +91,15 @@ <h2 class="govuk-heading-l">{{result.name}}</h2>
...
-                <tr class="govuk-table__row">
-                  <th scope="row" class="govuk-table__header">Table 3</th>
-                  <td class="govuk-table__cell">This is description about the table</td>
-                  <td class="govuk-table__cell">  <a href="#" class="govuk-link">Schema details</a></td>
-                </tr>
+                {% for table in tables %}
+                {% with table_type=table.type|lower %}
+                  <tr class="govuk-table__row">
+                    <th scope="row" class="govuk-table__header">{{table.name}}</th>
+                    <td class="govuk-table__cell">{{table.description}}</td>
```
MatMoore, line 98: I think this should be passed through the markdown filter so we can have rich text descriptions for tables as well as data products. What do you think?

home/service/details.py
```diff
@@ -1,24 +1,54 @@
 from data_platform_catalogue.search_types import MultiSelectFilter, ResultType
 from django.core.exceptions import ObjectDoesNotExist
 
+
 from .base import GenericService
 
 
-class DetailsService(GenericService):
```
murdo-moj, line 7: `DataProductDetailsService` reads better to me

tests/conftest.py
```diff
@@ -70,7 +79,12 @@ def mock_catalogue():
         mock_catalogue, page_results=generate_page(), total_results=100
     )
     mock_search_facets_response(mock_catalogue, domains=generate_options())
-
+    mock_list_data_product_response(
+        mock_catalogue,
+        page_results=generate_page(page_size=1, result_type=ResultType.TABLE),
+        total_results=1,
+    )
+    # mock_get_dataproduct_aspect(mock_catalogue)
```
MatMoore, line 87: commented out line here

tests/conftest.py
```diff
@@ -112,10 +144,15 @@ def search_context(valid_form):
 
 
 @pytest.fixture
-def detail_context(mock_catalogue):
+def detail_dataproduct_context(mock_catalogue):
     mock_catalogue.search.return_value = SearchResponse(
         total_results=1, page_results=generate_page(page_size=1)
     )
-    details_service = DetailsService(urn="urn:li:dataProduct:test")
+    # mock_catalogue.list_data_product_assets.return_value = SearchResponse(
```
MatMoore, line 151: can this mock value be removed now?

home/service/details.py
```diff
@@ -1,24 +1,54 @@
...
         filter_value = [MultiSelectFilter("urn", [urn])]
-        search_results = self.client.search(
-            query="", page=None, filters=filter_value)
+        search_results = self.client.search(query="", page=None, filters=filter_value)
 
         if not search_results.page_results:
             raise ObjectDoesNotExist(urn)
 
         self.result = search_results.page_results[0]
+        self.data_product_name = self.result.name
```
murdo-moj, line 20: this only gets used once, can you just reference `self.result.name`? or rename `self.result` to `self.data_product_search_result` maybe?

home/service/details.py
```diff
@@ -1,24 +1,54 @@
...
-class DetailsService(GenericService):
+class DetailsDataProductService(GenericService):
     def __init__(self, urn: str):
         self.urn = urn
         self.client = self._get_catalogue_client()
 
         filter_value = [MultiSelectFilter("urn", [urn])]
-        search_results = self.client.search(
-            query="", page=None, filters=filter_value)
+        search_results = self.client.search(query="", page=None, filters=filter_value)
```
murdo-moj, line 14:
```suggestion
        search_results = self.client.search(query="", page=None, filters=filter_value)
        assert len(search_results.page_results) <= 1
```

we should only ever get back 1 or 0 results right?

templates/details_data_product.html
```diff
@@ -82,7 +82,7 @@ <h2 class="govuk-heading-l">{{result.name}}</h2>
             </div>
             <br/>
             <table class="govuk-table">
-              <caption class="govuk-table__caption govuk-table__caption--m">Database content</caption>
+              <caption class="govuk-table__caption govuk-table__caption--m">Data Product content</caption>
```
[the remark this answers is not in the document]
LavMatt: Have matched the case used in search result i.e. `Data product`

home/service/details.py
LavMatt: Only ever passed a single urn and since these are unique will only get back single results

murdo-moj (approved):
Selenium tests are broken but we'll fix this in a later commit

murdo-moj (approved):

PR #96, opened 2024-02-26
Find moj data 69/glossary

murdo-moj:
* Added glossary view
* Added glossary service
* Added glossary template
* All of these will account for one parent node, or no parent node.


home/service/glossary.py
```diff
@@ -0,0 +1,46 @@
...
+from django.core.exceptions import ObjectDoesNotExist
+from copy import deepcopy
+from itertools import groupby
+
+from .base import GenericService
+
+
+class GlossaryService(GenericService):
+    def __init__(self):
+        # Can we put the client instantiation in the base class?
```
MatMoore, line 11: Sounds like a good idea to me. Either that or create a new base class especially for services that uses the catalogue.

home/service/glossary.py
```diff
@@ -0,0 +1,46 @@
...
+        self.client = self._get_catalogue_client()
+        self.context = self._get_context()
+
+    def _get_context(self):
+        """Returns a glossary context which is grouped by parent term"""
+        glossary_search_results = self.client.get_glossary_terms()
+        total_results = glossary_search_results.total_results
+
+        # The sort is required for the grouping to work correctly
+        page_results_copy = deepcopy(glossary_search_results.page_results)
```
MatMoore, line 21: If you use `sorted(list)` rather than `list.sort` it will create a copy as part of the sort, so you might not need this line(?)

MatMoore (approved):
Haven't checked the design itself but code changes look good.

Is there a figma for this page we should be comparing to?

I noticed we don't the sticky sidebar yet but I think it's worth doing that as a separate task as we need to make sure the layout is still responsive to different screen sizes.

home/service/glossary.py
murdo-moj: I will apply this

MatMoore (approved):

PR #136, opened 2024-03-07
Introduce table details page

MatMoore:
This PR populates part of the dataset details page to match the mockup in figma: 

https://www.figma.com/file/zTOVhHRnWfxRJqH6Gr8cIc/Catalogue-sketches?type=design&node-id=415-3214&mode=design&t=a6U0gCNNuShQRy41-0

- Breadcrumb (this is ported to the data product details as well)
- Title and description
- Column details

With this in place we can navigate to and from the search page, data product details and dataset details.

https://github.com/ministryofjustice/find-moj-data/assets/87579/148ae076-9e0a-4498-bc6f-2383449f1b93

## Design notes
There is no component for the grey box in the design, so I had to introduce some custom css for this:

```css
.summary-card {
    background-color: govuk-colour("light-grey");
    padding: govuk-spacing(3) govuk-spacing(4);
    margin-bottom: govuk-spacing(6);
}
```

On the data product details page we are using the card variant of the [summary list](https://design-system.service.gov.uk/components/summary-list/) and applying the card title class to everything to the whole thing to get the grey shading, but that doesn't work here, because we need the card to contain both a summary list and additional copy for the name and description.

Since we are preprocessing in sass this can directly reference the GOV.UK palette and spacing scale so this is easy to keep consistent with the design system.

## Out of scope for this PR
The following is left for future PRs:
- Metadata below the description
- Owner contact details on the right hand side
- "Resources" section (we haven't discussed this, so I think we can create a separate ticket for this one)

Additionally, I think it would be useful include [tags](https://design-system.service.gov.uk/components/tag/) to highlight primary keys and perhaps foreign keys in the schema. I left this out for now, because we don't have any metadata in datahub that includes it (neither the dbt import nor the data discovery tool import picked up primary keys), but if the schema does have this information, the data catalogue library will include this so we can do something with it in the frontend later.


mitchdawson1982 (approved):
Looks good

PR #160, opened 2024-03-11
Show contact info and metadata on all details pages

MatMoore:
Figma: https://www.figma.com/file/zTOVhHRnWfxRJqH6Gr8cIc/Catalogue-sketches?type=design&node-id=203-2121&mode=design&t=fR5PQmBviZV30T0v-0

This PR extracts a partial for the contact information box shared by the data products details and dataset details. I've also adapted it to look more like the design, although this will need further work to define the different roles in datahub and make sure they are returned correctly to the frontend.

Currently we are only fetching the data owner from the datahub API, but we should also show a point of contact for access requests. I've defaulted to using the data owner for this as well, but we will need to do some more work to define the various roles in datahub.

I've added some of the metadata fields to the details page but not all of it is available in the backend yet. I will either raise separate tickets for this or do it in follow up PRs.

I'm not sure yet what this should look like when data is incomplete - for now I've just left stuff blank as we do on the data product details page.

## Example data product with a data owner set
Screenshot 2024-03-11 at 16 43 45

## Example data product without a data owner set
Screenshot 2024-03-11 at 16 52 35

## Example dataset without a data owner set
Screenshot 2024-03-11 at 16 54 52


tom-webber (approved):
LGTM

PR #162, opened 2024-03-12
Lint django templates

MatMoore:
Introduce https://www.djlint.com/ for formatting and linting templates, as suggested by @tom-webber


MatMoore:
I'm a bit wary of relying on this linter actually, as it's quite aggressive in reformatting templated blocks, and can at times break the indenting, report false positives, and introduce unwanted whitespace

I'd rather not adopt this before the following issues are fixed:
https://github.com/djlint/djLint/issues/483
https://github.com/djlint/djLint/issues/636

For now I'm going to close this, and try https://github.com/rtts/djhtml instead. This doesn't do the linting, but does give us the benefit of consistently formatted templates.

PR #170, opened 2024-03-13
Condensed metadata component

MatMoore:
## Description

This PR condenses the metadata displayed on both details pages, so there is less whitespace between items.

Rather than using the GOV.UK summary list component, we're just using a regular list.

Figma: https://www.figma.com/file/zTOVhHRnWfxRJqH6Gr8cIc/Catalogue-sketches?type=design&node-id=415-3214&mode=design&t=UQKg4FYZpaX5nC6P-0

The second commit adds in the correct dataset name, as previously it was hardcoded.

## Before

Screenshot 2024-03-13 at 16 50 46
Screenshot 2024-03-13 at 16 51 15

## After

Screenshot 2024-03-13 at 16 44 17
Screenshot 2024-03-13 at 16 50 15

### DIFF
--- a/templates/details_data_product.html
+++ b/templates/details_data_product.html
@@ -31,47 +31,36 @@
       <div class="govuk-body">
           {{result.description|markdown:3}}
       </div>
-      <dl class="govuk-summary-list govuk-summary-list--no-border">
-        <div class="govuk-summary-list__row">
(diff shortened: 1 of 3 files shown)


murdo-moj (approved):
LGTM

PR #174, opened 2024-03-14
Swapped dev and test environments

murdo-moj:
* Test is for testing on users
* Dev means development for developers
* So we should deploy new changes to dev first then test


.github/workflows/test.yml
tom-webber, line 1: rename file and workflow to 'tests and deploy'

.github/workflows/test.yml
```diff
@@ -82,19 +82,19 @@ jobs:
       - name: Run javascript tests
         run: npm test
 
-  # Only the deploy workflow should deploy to the test env on main
+  # Only the deploy workflow should deploy to the dev env on main
```
tom-webber, line 85:
```suggestion
  # Deploys to the 'dev' frontend unless pushed to main
```

tom-webber (changes requested):

.github/workflows/test.yml
murdo-moj: the idea was that doing a deploy to dev is sometimes part of the test workflow you do - can I just leave the file name as is?

tom-webber: test is a confusing file (and workflow) name for something that deploys to dev

tom-webber (approved):

Issue #153, opened 2024-03-11
Display environment in front end


seanprivett:

murdo-moj:
https://docs.djangoproject.com/en/5.0/ref/templates/api/#writing-your-own-context-processors I added a context processor which adds settings to the context of every view; this is Django's preferred way.

PR #183, opened 2024-03-14
Added environment name to alpha banner

murdo-moj:
https://github.com/ministryofjustice/find-moj-data/issues/153
- I added the environment string as a setting then to each view via a context processor, which Django suggest as a way of making variables available to all templates https://docs.djangoproject.com/en/5.0/ref/templates/api/#writing-your-own-context-processors


LavMatt (approved):

mitchdawson1982 (approved):
LGTM

Issue #156, opened 2024-03-11
update data product labels on front end to align with our new definition in DataHub


seanprivett:
follows on from https://github.com/ministryofjustice/find-moj-data/issues/175

PR #205, opened 2024-03-21
Dc 156 add database entities

LavMatt:
This PR relates to https://github.com/ministryofjustice/find-moj-data/issues/156

Front end now recognises Databases (datahub container entities). The changes made should still work with the dev and test datahub instances where we have the metadata ingested classified different and a different set of domains.

It makes use of `ministryofjustice-data-platform-catalogue v0.22.0` which builds in additional entities and methods to get the parents and children of datasets (both data products and containers) and containers.


templates/details_dataset.html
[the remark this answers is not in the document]
LavMatt: Quite possibly yeah. Database is a subtype of container (and that's the name used for that template) so we should probably align and do the same for both entity types Dataset/Table Container/Database

home/service/details.py
```diff
@@ -56,40 +57,94 @@ def _get_context(self):
...
+
+        entities_in_database = sorted(entities_in_database, key=lambda d: d["name"])
+
+        return entities_in_database
+
+    def _get_context(self):
+        context = {
+            "result": self.result,
+            "result_type": (
+                "Database"
```
murdo-moj, line 101: isn't the result_type always database here?

home/service/search.py
```diff
@@ -25,7 +25,7 @@ def domains_with_their_subdomains(domain: str, subdomain: str) -> list[str]:
 
     subdomains = DomainModel().subdomains.get(domain, [])
     subdomains = [subdomain[0] for subdomain in subdomains]
-    return [domain, *subdomains] if subdomains else []
+    return [domain, *subdomains] if not domain == "" else []
```
murdo-moj, line 28:
```suggestion
    return [domain, *subdomains] if domain != "" else []
```

templates/details_database.html
```diff
@@ -0,0 +1,104 @@
...
+
+  <div class="govuk-grid-column-one-third">
+    {% include "partial/contact_info.html" with data_owner=result.metadata.owner data_owner_email=result.metadata.owner_email %}
+  </div>
+</div>
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-full">
+    <table class="govuk-table">
+      <caption class="govuk-table__caption govuk-table__caption--m">Data product content</caption>
```
murdo-moj, line 74:
```suggestion
      <caption class="govuk-table__caption govuk-table__caption--m">Database content</caption>
```

home/service/details.py
LavMatt: Yeah i think it's not needed and was accommodating dodgy test inputs rather than actually being of any use.

LavMatt: have committed fix

murdo-moj (approved):

PR #207, opened 2024-03-22
Added entity type search filter

murdo-moj:
- Added entity type filter
- Removed classification filter

### DIFF
--- a/home/forms/search.py
+++ b/home/forms/search.py
@@ -35,10 +35,10 @@ def get_where_to_access_choices():
 
 
 def get_entity_types():
-    return [
+    return sorted([
         (entity.name, entity.name.replace("_", " ").lower().title())
         for entity in ResultType if entity.name != "GLOSSARY_TERM"
-    ]
+    ])
 
 
 class SelectWithOptionAttribute(forms.Select):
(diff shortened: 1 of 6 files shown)


home/forms/search.py
```diff
@@ -29,16 +30,17 @@ def get_sort_choices():
...
-
 def get_where_to_access_choices():
     return [("analytical_platform", "Analytical Platform")]
 
 
+def get_entity_types():
+    return [
+        (entity.name, entity.name.replace("_", " ").lower().title())
+        for entity in ResultType if entity.name != "GLOSSARY_TERM"
+    ]
```
LavMatt, line 41: could we sort these maybe? Something like
```suggestion
    entities = [
        (entity.name, entity.name.replace("_", " ").lower().title())
        for entity in ResultType
        if entity.name != "GLOSSARY_TERM"
    ]
    entities.sort(key=lambda x: x[0])
    return entities

LavMatt (commented):
Looks good. One minor take it or leave it comment but also should change line 127 in search service to `label_clear_href["Where To Access"] = where_to_access_clear_href` as it still displays `Availability` for the filter label (will probs need to change some tests to accoutn for that too)

LavMatt (approved):

Issue #198, opened 2024-03-20
Make table captions descriptive


jemnery:
It's more helpful for a screen reader if table captions are specific. This needs to balance with page design - not making captions too wide

"Schema" could be "Table schema"

![Image](https://github.com/ministryofjustice/find-moj-data/assets/1306705/658363f7-25d1-45c5-a906-f7f191b28ba2)

PR #243, opened 2024-04-09
Improve captions on details pages

MatMoore:
Resolves https://github.com/ministryofjustice/find-moj-data/issues/198

Screenshot 2024-04-09 at 13 39 12

This PR also makes a few other small tweaks:
- Fix the heading structure so we have an h1 on these pages
- Use "table" rather than "schema" as the link text when linking to the table details
- Don't render a table of tables if there are no tables

### DIFF
--- a/templates/details_data_product.html
+++ b/templates/details_data_product.html
@@ -91,7 +91,7 @@
                       {{ table.description|markdown:3 }}
                     {% endif %}
                   </td>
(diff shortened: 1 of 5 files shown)


jemnery:
Excellent 👍

jemnery (approved):

jemnery (approved):

PR #246, opened 2024-04-10
Workaround for skip link issues

MatMoore:
I forgot to commit this yesterday, but this is a workaround for skip links not being underlined. I've also reported the issue in the govuk-design-system slack channel.

The $govuk-global-styles setting applies the [govuk-focus-text mixin](https://github.com/alphagov/govuk-frontend/blob/bd465bda9bea6b46aee8ac8f15b607a3b51c2b3b/packages/govuk-frontend/src/govuk/helpers/_focused.scss#L25) to all links, which removes underline and adds box-shadow.  But the skip link component [removes box shadow from the focus state](https://github.com/alphagov/govuk-frontend/blob/main/packages/govuk-frontend/src/govuk/components/skip-link/_index.scss#L30), without adding back in the underline.  This has the effect that skip links are not underlined when $govuk-global-styles is true.

This workaround reenables the underline. This commit may be reverted when we pull a new version of the design system with a fix in.

Before:

Screenshot 2024-04-10 at 10 24 49

After:

Screenshot 2024-04-10 at 10 23 31

Component docs: https://design-system.service.gov.uk/components/skip-link/


tom-webber (approved):
LGTM

Issue #223, opened 2024-03-28
🐛 Search query results not always displaying matched descriptions


LavMatt:
### Describe the bug.

In the search results page, because we trim descriptions, we do not display the matched string from the search query in the description if it occurs after the cut-off.

### To Reproduce

search `escapes` in fmd front end.  A releases in error chart is returned. The match is description but you'll see the term `escapes` is not highlighted in the description because we've trimmed it before this part is shown

### Expected Behaviour

you’d expect to see the section of the description where your search query matched and the matched string highlighted, in all search results.

MatMoore:
Related to this, if you have multiple words in the query, only the full match will be highlighted. We could improve this by splitting on words and highlighting each word match independently.

PR #259, opened 2024-04-15
Better highlighting in descriptions

MatMoore:
This PR extends the previous code for highlighting search terms on the results page. The goal is to show the matched terms in context, like what you would see in google search results.

This should mostly resolve issues where snippets are being truncated too early, hiding the search terms:
https://github.com/ministryofjustice/find-moj-data/issues/223

## New behaviour
This is a combination of multiple changes.

Although this repeats some of the processing the search engine does internally, I think it's useful to do this here because the GraphQL API does not provide a way to return matched snippets, but these snippets are still much better than what we can get from naively truncating the description.

### Highlight terms that match part of the query
Instead of applying `<mark></mark>` tags around all instances of the search term, we now do some processing on the search query to come up with a set of terms to highlight. The algorithm now looks like this:

1. Split the search term on spaces to get individual words
2. Stem each word 
3. For each term, and the search query as a whole, look for matches at the start of any words and apply mark tags around the matching words as before

This means we can search for "escapes" and get "escape" or "escaping" highlighted.

### Increase snippet size
I've also increased the snippet size - so the problem is less likely to occur - but I can revert this change if we prefer the shorter snippets.

### Custom template filter instead of string truncation
Finally, I've written a custom template filter to create search snippets based on where the `<mark></mark>` tags are. Previously, we just truncated after a certain number of characters, which also left invalid HTML syntax when the mark was truncated in the middle.

I've implemented a number of heuristics to improve this:
1. Base all truncation *at the end of the text* on django's `truncatechars_html` - this is HTML aware, and closes any unclosed tags
2. If possible, drop any preceding paragraphs, until we have one with marked terms
3. If necessary, drop words at the beginning of the paragraph, so that the marked text doesn't get truncated. Leave enough words so that the sentence still makes sense.

## Deployment changes
Since I've introduced nltk as a dependency, we need to download the data for the stemmer to use.

To do this I added the following line to the dockerfile:

```
RUN poetry run python -m nltk.downloader punkt
```

## Out of scope
This PR focuses on technical improvements we can make so we show the best thing even if we have bad content. However, we can also look at improving the content itself, for example by:

- frontloading keywords where possible
- making sure we strip unnecessary html (such as `<p class="govuk-body">` when scraping metadata from justice data and other sources
- manually overriding the search summaries via custom properties (as we did before for data products)

## Before
Snippet from the start of the text with no marked terms

## With new truncation filter and original snippet size
Screenshot 2024-04-15 at 10 11 54

## With new filter and longer snippet size
Snippet from the middle of the text, with marked terms


LavMatt (approved):
That's much better and like how we now it highlights words where search results have be been returned on non-exact matches, e.g search "escapes" and get back "escaping" highlighted in description.

I think increase to 300 is OK, and is room for enough context before and after search matches. Seeing as we plan to trim the standard metadata fields displayed for a search result, shouldn't ultimately result in space lost

PR #260, opened 2024-04-15
Fmd 217 error pages

murdo-moj:
* Added 404 error template with gov.uk content
* Added 500 error template (this is what gets triggered as a catch-all for all datahub errors at the moment)
* Added 400 error template

### DIFF
--- a/templates/500.html
+++ b/templates/500.html
@@ -11,6 +11,9 @@
           <p class="govuk-body">
             There is a problem with the server.
           </p>
+          <p class="govuk-body">
+            If the server problems persist, <a href="https://moj.enterprise.slack.com/archives/C03QZ776JVA" class="govuk-link">contact the find-moj-data team on Slack.</a>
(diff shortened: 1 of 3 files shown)


templates/500.html
```diff
@@ -0,0 +1,19 @@
...
+
+{% block content %}
+
+  <div class="govuk-width-container">
+    <main class="govuk-main-wrapper govuk-main-wrapper--l" id="main-content" role="main">
+      <div class="govuk-grid-row">
+        <div class="govuk-grid-column-two-thirds">
+          <h1 class="govuk-heading-l">Server error</h1>
+          <p class="govuk-body">
+            There is a problem with the server.
```
MatMoore, line 12: Maybe we could link to our slack channel as well - like if the problem persists let us know here

MatMoore (approved):

MatMoore (approved):

Issue #240, opened 2024-04-05
🐛 Table column descriptions are fully displayed irrespective of their length


LavMatt:
### Describe the bug.

We are not applying any limitations to the default number of characters shown for column descriptions. Unlike other trimmed descriptions there is not another page where these descriptions cans be displayed more fully in isolation

Nevertheless when column descriptions are lengthy they can overwhelm the schema table.

### To Reproduce

have a look at the `xhibit_derived.all_offence` table in find-moj-data

### Expected Behaviour

Descriptions should be cut-off at some limit as default, with the option for a user to expand to see the full description

MatMoore:
I don't think there is a good design system component for this, but another team implemented something similar here: https://github.com/alphagov/govuk-design-system-backlog/issues/44#issuecomment-1674995951

[...]

PR #261, opened 2024-04-15
Limit column descriptions temporarily

MatMoore:
This is a partial workaround for https://github.com/ministryofjustice/find-moj-data/issues/240

A more flexible solution would be to provide a "Show more" button, but I'm going to look at that separately, as I'm not sure how much work it would be to implement.

In some cases we have very long descriptions associated with columns. Datahub has no restrictions on what we can store here, however our layout has some issues with long text:

1. It is hard to scan through the column list if a single column takes up the whole page
2. If a column description contains a markdown table, we don't render that as a table inside the table
3. If we have unparsed markdown with no natural line breaks (such as a column header) then the table grows to encompass its width

This workaround allows the problem page to render properly until these problems are addressed.

### DIFF
--- a/templates/details_table.html
+++ b/templates/details_table.html
@@ -89,7 +89,7 @@
             {% for column in table.column_details %}
               <tr class="govuk-table__row">
                 <td class="govuk-table__cell">{{column.name}}</td>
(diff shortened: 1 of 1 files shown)


hjribeiro-moj (approved):
LGTM

PR #267, opened 2024-04-16
Stable tables

MatMoore:
A few more workarounds to avoid bad content from messing up tables, as can be seen on this page: https://data-platform-find-moj-data-test.apps.live.cloud-platform.service.justice.gov.uk/details/table/urn:li:dataset:(urn:li:dataPlatform:athena,xhibit_derived.all_offence,PROD)

This happens because tables are sized according to their contents, so if you have any cell that is wide, it will expand the table to accommodate it.

The limit is a bit arbitrary, and I've skipped smaller breakpoints for now because these tables do not render nicely either way on small screens.

This PR compliments https://github.com/ministryofjustice/find-moj-data/pull/261

## Without max-size
Screenshot 2024-04-16 at 16 13 22

## With max-size
Screenshot 2024-04-16 at 16 14 39


hjribeiro-moj (approved):
LGTM

Issue #248, opened 2024-04-11
Spike: Search doesn't perform very well with multiple words


seanprivett:
- Currently, search on datahub with multiple words matches the whole search string exactly to results
- This means that any typo or error in word order results in few or zero search results
- This restricts the practical use of the search functionality to 2 or 3 words
---
- DataHub offers some flexibility in how it's search works
- Currently the exclusive match setting is set to false, yet we can see that the search is exclusive. We've asked datahub about this on Slack, but have no response yet
---
- What do we want search behaviour to be like? We don't really know.

YvanMOJdigital:
spike to investigate options for improving search results when using multiple words

MatMoore:
I started investigating this last week and wrote some notes here: https://mojdt.slack.com/archives/C03QZ776JVA/p1712233855118319

murdo-moj:
Pasting Mat's notes from slack:

```Dug into this a bit more. This is our current config, with no customisations:
    ## Search related configuration
    search:
      ## Maximum terms in aggregations
      maxTermBucketSize: 20

      ## Configuration around exact matching for search
      exactMatch:
        ## if false will only apply weights, if true will exclude non-exact
        exclusive: false
        ## include prefix exact matches
        withPrefix: true
        ## boost multiplier when exact with case
        exactFactor: 2.0
        ## boost multiplier when exact prefix
        prefixFactor: 1.6
        ## stacked boost multiplier when case mismatch
        caseSensitivityFactor: 0.7
        ## enable exact match on structured search
        enableStructured: true
What's interesting to me is it has exclusive: false, so non-exact matches are not supposed to be excluded
Here is a test case demonstrating that non-exact matches are excluded:
query {
  searchAcrossEntities(
    input: {
      query: "An abscond is an escape that does not involve overcoming a physical security restraint"
    }
  ) {
    total
  }
}
-> match
query {
  searchAcrossEntities(
    input: {
      query: "An abscond is an escape that does not involve overcoming a gorilla"
    }
  ) {
    total
  }
}
-> does not match
[The same behaviour is visible in the datahub UI, so it doesn't look like something we've introduced.](https://datahub-catalogue-preprod.apps.live.cloud-platform.service.justice.gov.uk/search?page=1&query=An%2520abscond%2520is%2520an%2520escape%2520that%2520does%2520not%2520involve%2520overcoming%2520a%2520gorilla&unionType=0)
[...]

PR #293, opened 2024-04-30
Add a details component with search query tips

MatMoore:
Provide some query examples highlighting some of the supported syntax, and link to the full documentation in Datahub.

This should help mitigate the sometimes unintuitive default behaviour of requiring that ALL search terms match (https://github.com/ministryofjustice/find-moj-data/issues/248)

This behaviour is how Datahub have configured search, and to change it we would have to customise the Elasticsearch query.

This copy has not been reviewed by a content designer yet - it will probably need another iteration.

### Collapsed
Screenshot 2024-04-30 at 15 08 40

### Expanded
Screenshot 2024-04-30 at 15 08 46

### DIFF
--- a/templates/search.html
+++ b/templates/search.html
@@ -21,6 +21,22 @@
           </div>
         </div>
       </form>
+      <details class="govuk-details">
+        <summary class="govuk-details__summary">
+          <span class="govuk-details__summary-text">
+            Search query tips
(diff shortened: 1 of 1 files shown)


hjribeiro-moj (approved):
LGTM

PR #308, opened 2024-05-03
:technologist: Integrate front end with Sentry

mitchdawson1982:


core/settings.py
```diff
@@ -168,3 +169,17 @@
         },
     },
 }
+
+# Sentry Configuration
+sentry_sdk.init(
+    # dsn="https://1936689219fbec45b1437ee54bcd2a29@o345774.ingest.us.sentry.io/4507181591101440",
```
MatMoore, line 175: I think this should be removed from the commit if it allows anyone to publish to our sentry?

core/settings.py
```diff
@@ -168,3 +169,16 @@
...
+
+# Sentry Configuration
+sentry_sdk.init(
+    dsn=os.environ.get("SENTRY_DSN"),
+    # Set traces_sample_rate to 1.0 to capture 100%
+    # of transactions for performance monitoring.
+    traces_sample_rate=1.0,
+    # Set profiles_sample_rate to 1.0 to profile 100%
+    # of sampled transactions.
+    # We recommend adjusting this value in production.
```
MatMoore, line 181: Can you allow these to be overriden by environment variables so we can adjust for prod later?

MatMoore (approved):

core/settings.py
mitchdawson1982: Yes, I am happy to change this when required for Prod.

PR #319, opened 2024-05-08
Fix rendering of some details pages

MatMoore:
Resolves https://github.com/orgs/ministryofjustice/projects/57/views/5?pane=issue&itemId=61705328#:~:text=not%20always%20parseable-,%23303,-Edit

Fixes on this branch:

1. Fix links to the parent entity in breadcrumbs (id attribute changed to URN)
2. Make sure selenium tests call out to an up to date version of axe-core, instead of using the same cached version forever
3. Remove incorrect validation on Column.name, and use Column.display_name instead for rendering
4. Do not render timestamps of '0' (January 1st 1970)


hjribeiro-moj (approved):
LGTM

PR #320, opened 2024-05-08
Tidy empty search results page

MatMoore:
Related to https://github.com/orgs/ministryofjustice/projects/57/views/5?pane=issue&itemId=60708103 we want to gracefully handle the case where the search returns no results.

In this case there is no point rendering a "1" from the pagination, and we can skip the sort controls as well. I've introduced a partial so we can show a helpful message here. Given our research participants struggled with long queries in particular, we can nudge them towards shorter queries here (this will change if we later customise the search configuration)

In the process I also fixed a couple of bugs:
- I noticed than when we have > 10,000 results we just get 10,000 back, so I've formatted this as "10,000+"
- The "Matched fields" value was breaking when processing some custom properties set via DBT

Screenshot 2024-05-08 at 13 51 28
Screenshot 2024-05-08 at 14 11 28
Screenshot 2024-05-08 at 14 12 17

### DIFF
--- a/core/settings.py
+++ b/core/settings.py
@@ -140,6 +140,7 @@ DATABASES = {
 SERVICE_NAME = "Find MOJ Data"
 GOV_UK_SUFFIX = "GOV.UK"
 
+MAX_RESULTS = 10_000
 
 LOGGING = {
     "version": 1,  # the dictConfig format version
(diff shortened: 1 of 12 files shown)


hjribeiro-moj (dismissed):
LGTM

hjribeiro-moj (approved):
LGTM

PR #330, opened 2024-05-10
Fmd-314-automate-static-build

murdo-moj:
Automate static files being copied from node modules. We have created a few js files which aren't included in the build.


hjribeiro-moj (approved):
LGTM

.gitignore
```diff
@@ -162,7 +162,9 @@ cython_debug/
 .DS_Store
 
 #static files dirs
-staticfiles/
+/static
```
MatMoore, line 165: `staticfiles/` is still confiugured as the STATIC_ROOT, so we still need it here

`STATIC_ROOT = BASE_DIR / "staticfiles"`

Dockerfile
```diff
@@ -3,7 +3,7 @@ FROM node:22-bullseye as node_builder
 WORKDIR /app
 COPY . .
 
-RUN npm install && npm run sass
+RUN npm install
```
MatMoore, line 6: Do you not need to run the new `npm run dependencies` command for this to be equivalent?

MatMoore: Ah, nevermind, I see this is automatically run 👍🏻

murdo-moj: dependencies is automatically triggered after npm makes any changes to the node_modules folder, so after any install or most updates

hjribeiro-moj (approved):

PR #332, opened 2024-05-13
Add google analytics JS

MatMoore:
This is conditional on an environment variable being set.

Once we have a cookie banner / cookie page, we can also skip rendering the javascript if analytics cookies are not accepted.


core/settings.py
```diff
@@ -171,6 +174,8 @@
     },
 }
 
+ENABLE_ANALYTICS = os.environ.get("ENABLE_ANALYTICS") in TRUTHY_VALUES
```
tom-webber, line 177:
```suggestion
ENABLE_ANALYTICS: bool = os.environ.get("ENABLE_ANALYTICS", "") in TRUTHY_VALUES
```

templates/base/base.html
```diff
@@ -60,5 +60,9 @@
     </script>
     {% block scripts %}
     {% endblock scripts %}
+    {% if enable_analytics %}
+    <script async src="https://www.googletagmanager.com/gtag/js?id={{analytics_id}}"></script>
+    <script>window.dataLayer = window.dataLayer || []; function gtag(){dataLayer.push(arguments);} gtag('js', new Date()); gtag('config', '{{analytics_id}}');</script>
+    {% endif %}
```
tom-webber, line 66: ⛏️ minor suggestion for emphasising environment variable usage
```suggestion
    <script async src="https://www.googletagmanager.com/gtag/js?id={{ ANALYTICS_ID }}"></script>
    <script>window.dataLayer = window.dataLayer || []; function gtag(){dataLayer.push(arguments);} gtag('js', new Date()); gtag('config', '{{ ANALYTICS_ID }}');</script>
    {% endif %}
```

core/settings.py
```diff
@@ -171,6 +174,11 @@
     },
 }
 
+ANALYTICS_ID: str = os.environ.get("ANALYTICS_ID", "")
+ENABLE_ANALYTICS: bool = (
+    os.environ.get("ENABLE_ANALYTICS") in TRUTHY_VALUES
+) and ANALYTICS_ID != ""
```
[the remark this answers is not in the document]
MatMoore: This isn't needed, `.get` defaults to None which is not in the list

templates/base/base.html
MatMoore: this needs to match the context processor. I've kept it lowercase to be consistent with how we've done the env variable

```
{"env": settings.ENV}
```

tom-webber (approved):
Nice!

PR #333, opened 2024-05-13
Fmd 305 output prometheus metrics

LavMatt:
adds `django-prometheus` to fmd, creating a metrics endpoint we can pull into cloud platform prometheus


deployments/templates/service.yml
```diff
@@ -2,6 +2,9 @@ apiVersion: v1
 kind: Service
 metadata:
   name: find-moj-data-service
+  namespace: data-platform-find-moj-data-dev
```
tom-webber, line 5:
```suggestion
  namespace: ${NAMESPACE}
```

deployments/templates/monitoring/fmd-servicemonitoring.yml
tom-webber, line 1: ⛏️ I would rename this file to `fmd-servicemonitor.yml` to follow k8s naming convention

core/settings.py
```diff
@@ -23,6 +24,7 @@
 DEBUG: bool = DEBUG_STR in TRUTHY_VALUES
 
 ALLOWED_HOSTS = str(os.environ.get("DJANGO_ALLOWED_HOSTS")).split(" ")
+ALLOWED_HOSTS.append(gethostbyname(gethostname()))
```
murdo-moj, line 27: https://stackoverflow.com/questions/37031749/django-allowed-hosts-ips-range I see a suggestion on this stack page on an improvement on this for docker. did you try that and discount it?

murdo-moj: > Update. If you are using docker the following code is better as gethostbyname doesn't get the correct information.
```
from socket import gethostname, gethostbyname, gethostbyname_ex
ALLOWED_HOSTS = [ gethostname(), ] + list(set(gethostbyname_ex(gethostname())[2]))
```

tom-webber: [gethostbyname_ex()](https://pythontic.com/modules/socket/gethostbyname_ex) doc for context

LavMatt: I did not try that as what i did worked but if there a risk of it not always doing so i can try this

murdo-moj (approved):

core/settings.py
LavMatt: So looked into this in a bit more detail. The main difference between `gethostbyname` and `gethostbyname_ex` is the former returns a string of the host ip address and the later returns a tuple of 3 items where the last is a list of host ip addresses. 

So if we needed to add more than one host ip address this would be what to use but as we only have a single one i think we're good with whats been used

Issue #327, opened 2024-05-09
🐛 Some values may be null in datahub, but are non-null in our models


MatMoore:
### Describe the bug.

https://ministryofjustice.sentry.io/issues/5321294599/?alert_rule_id=15157261&alert_type=issue&notification_uuid=ccc3a005-8ade-4aeb-8b87-c209758db688&project=4507181591101440&referrer=slack

In this case description is None, but we are validating that description is a string.

### Expected Behaviour

Two options

1. We change our model to match datahub
2. Cooerce None values to string when reading from datahub. Anytime we do `.get(field, "")` on the graphql response we need to make sure we handle any case where the field is set, but set to None

PR #337, opened 2024-05-14
Edit parse properties to convert some None values to

hjribeiro-moj:
🐛 fix for https://github.com/ministryofjustice/find-moj-data/issues/327

- updated `parse_properties` to look for fields requiring default of "" and convert it if filed was explicitly set to `None`. Found 2 examples: `descritpion` and `externalUrl`.
- Added a unit test case for this
- Added test evidence of the fix working on the APP

### DIFF
--- a/lib/datahub-client/data_platform_catalogue/client/graphql_helpers.py
+++ b/lib/datahub-client/data_platform_catalogue/client/graphql_helpers.py
@@ -16,6 +16,8 @@ from data_platform_catalogue.entities import (
     UsageRestrictions,
 )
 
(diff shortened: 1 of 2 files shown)


hjribeiro-moj:
**Test Evidence**

before fix when searching for   "xhibit_all_offence_disposal_stg":
image

with fix:
image

LavMatt (commented):
LGTM - the other cases i was thinking of are actually already handled in the code, e.g relationships, i just was off during that part of the refactor

LavMatt (approved):

PR #342, opened 2024-05-15
DP-212 Azure auth configuration

mitchdawson1982:
### Additions
- Custom User model via AbstractBaseUser
- Custom User Manager via BaseUserManager
- Azure authentication via [django-azure-auth](https://github.com/AgileTek/django-azure-auth)
- Sqlite flat-file local database for user and session storage
- User model CRUD tests
- Pytest session scoped fixture to bypass auth for web tests
- New variables for local development .env.example

### Removes
- Signed cookie session engine
- In memory DB
- Initial sample search data (unrelated)
- Unused selenium tests (unrelated)


core/settings.py
```diff
@@ -129,14 +132,11 @@
...
 
-# session
-SESSION_ENGINE = "django.contrib.sessions.backends.signed_cookies"
 
-# Not actually used - Just required for LiveServerTestCase
 DATABASES = {
     "default": {
         "ENGINE": "django.db.backends.sqlite3",
-        "NAME": ":memory:",
+        "NAME": "db.sqlite3",
```
MatMoore, line 134: Seems like the reason we rely on a database here is because [the backend inherits from ModelBackend](https://github.com/Weird-Sheep-Labs/django-azure-auth/blob/4ca256d0a3487094fcee9315dfba0d73284a8fc7/azure_auth/backends.py#L6) rather than [RemoteUserBackend](https://docs.djangoproject.com/en/5.0/ref/contrib/auth/#django.contrib.auth.backends.RemoteUserBackend)

It's possible there is another library that doesn't implement it this way, but I think it's quite likely we will need a database anyway at some point, so I'm not overly concerned about adding in a database dependency.

If we go down this path, then I assume we will need a follow up PR/ticket to configure Postgres rather than SQLite, so that in cloud platform multiple instances can share the same session store.

core/settings.py
```diff
@@ -192,3 +192,19 @@
...
+    "CLIENT_SECRET": os.environ.get("CLIENT_SECRET"),
+    "REDIRECT_URI": os.environ.get("REDIRECT_URI"),
+    "SCOPES": ["User.Read"],
+    "AUTHORITY": os.environ.get("AUTHORITY"),
+    "LOGOUT_URI": "https://localhost:8000/logout",
+}
+LOGIN_URL = "/azure_auth/login"
+LOGIN_REDIRECT_URL = "/search"  # Or any other endpoint
+
+AUTHENTICATION_BACKENDS = ("azure_auth.backends.AzureBackend",)
```
MatMoore, line 203: I think we can probably swap this for an [AllowAllUsers backend](https://docs.djangoproject.com/en/5.0/ref/contrib/auth/#django.contrib.auth.backends.AllowAllUsersModelBackend) in tests

users/models.py
```diff
@@ -0,0 +1,24 @@
...
+
+from .managers import CustomUserManager
+
+
+class CustomUser(AbstractBaseUser, PermissionsMixin):
+    username = models.EmailField(_("email address"), unique=True)
+    email = models.EmailField(_("email address"), unique=True)
+    first_name = models.CharField(_("first name"), max_length=100)
+    last_name = models.CharField(_("last name"), max_length=100)
+    is_staff = models.BooleanField(default=False)
```
MatMoore, line 14: I don't think this makes sense for us - users will be staff by definition

users/migrations/0001_initial.py
```diff
@@ -0,0 +1,90 @@
...
+                ),
+                ("password", models.CharField(max_length=128, verbose_name="password")),
+                (
+                    "last_login",
+                    models.DateTimeField(
+                        blank=True, null=True, verbose_name="last login"
+                    ),
+                ),
+                (
+                    "is_superuser",
```
MatMoore, line 36: This field doesn't seem to match the current model?

core/settings.py
mitchdawson1982: Agreed, if we have more than one FE instance then we will need a shared DB.

users/models.py
mitchdawson1982: `is_staff` is a value that is used to determine if someone can log in to the Django admin panel.

users/migrations/0001_initial.py
mitchdawson1982: I have removed the create superuser manager method but it was in the initial migration.

core/settings.py
mitchdawson1982: This is conditional on the AZURE_AUTH_ENABLED now so we wont have a backend when we dont need it.

.env.example
```diff
@@ -2,7 +2,15 @@
...
 DJANGO_ALLOWED_HOSTS=localhost 127.0.0.1 [::1]
+DJANGO_LOG_LEVEL=INFO
 
 # Catalogue Variables
 CATALOGUE_TOKEN=mycatalogtoken
 CATALOGUE_URL=https://datahub-catalogue-dev.apps.live.cloud-platform.service.justice.gov.uk/api/gms
+
+# Azure Variables (Examples)
+AZURE_AUTH_ENABLED=true # any string value other than true to disable Azure authentication
+AZURE_CLIENT_ID=< obtain from onepassword >
```
MatMoore, line 13: I think in one of the other projects we had a nice script that generated the .env based by pulling all the secrets from secrets manager. It's not essential right now, but just wondering if we could do the same here so when someone joins the team they just have run one command for this?

Dockerfile
```diff
@@ -44,6 +44,7 @@ COPY --from=builder ${VIRTUAL_ENV} ${VIRTUAL_ENV}
 
 
 RUN python manage.py collectstatic --noinput
+RUN python manage.py migrate
```
MatMoore, line 47: I don't think the migrate should run as part of the docker build, should it?

I would expect this to happen as part of the kubernetes deployment, because we will be moving away from sqlite, and we won't want the database to be baked into the image. Once we externalise the db, this will be environment dependent and we won't have access to the right credentials at build time.

core/settings.py
```diff
@@ -201,3 +196,23 @@
     profiles_sample_rate=1.0,
     environment=ENV or "local",
 )
+
+# Enable / Disable Azure Auth
+if os.environ.get("AZURE_AUTH_ENABLED", "true") == "true":
```
MatMoore, line 201: We want to be careful here because we don't want to accidentally disable auth because of a misconfiguration. I think that checking for an explicit value of "false" is a better approach.

MatMoore (commented):
Looks good so far, just a few minor sessions

core/settings.py
```diff
@@ -201,3 +196,23 @@
...
+    # Adds the Azure Authentication middleware to the Django Authentication middleware
+    MIDDLEWARE.append(
+        "azure_auth.middleware.AzureMiddleware",
```
tom-webber, line 204: This disrupts [`"django_prometheus.middleware.PrometheusAfterMiddleware"`](https://github.com/ministryofjustice/find-moj-data/pull/342/files#diff-4392745abac424baa573e9c60413308af753eb912be194dfc224f1e71e81ef98R57), which [needs to be the final middleware entry](https://github.com/korfuri/django-prometheus?tab=readme-ov-file#quickstart)

.env.example
mitchdawson1982: Yes I am all in favour of a script that pulls secrets from secret manager and builds the .env dynamically. Perhaps introduce a Makefile at the same time. Ill create a ticket for it.

core/settings.py
mitchdawson1982: Makes sense, ill amend.

deployments/templates/deployment.yml
```diff
@@ -38,6 +38,14 @@ spec:
...
             - name: ANALYTICS_ID
               value: "${ANALYTICS_ID}"
+            - name: AZURE_AUTH_ENABLED
+              value: "${AZURE_AUTH_ENABLED}"
+            - name: AZURE_CLIENT_ID
+              value: "$AZURE_CLIENT_ID"
+            - name: AZURE_REDIRECT_URI
+              value: "$AZURE_REDIRECT_URI"
+            - name: AZURE_AUTHORITY
+              value: "$AZURE_AUTHORITY"
```
tom-webber, line 48:
```suggestion
            - name: AZURE_CLIENT_ID
              value: "${AZURE_CLIENT_ID}"
            - name: AZURE_REDIRECT_URI
              value: "${AZURE_REDIRECT_URI}"
            - name: AZURE_AUTHORITY
              value: "${AZURE_AUTHORITY}"
```
minor env var formatting consistency

tests/conftest.py
```diff
@@ -3,8 +3,22 @@
 from typing import Any
 from unittest.mock import MagicMock, patch
 
+from django.conf import settings
+
 import pytest
 from data_platform_catalogue.client.datahub_client import DataHubCatalogueClient
+from data_platform_catalogue.entities import (
```
tom-webber, line 10: Why are there 2 import statements for `data_platform_catalogue.entities`? Also, can these imports be alphabetical (available in the command palette in VSCode).

tests/users/models/test_user.py
```diff
@@ -0,0 +1,108 @@
+from django.contrib.auth import get_user_model
+from django.db.utils import IntegrityError
+
+import pytest
+
+
+@pytest.mark.django_db
+class TestUserModel:
+
```
tom-webber, line 9:
```suggestion
    """
    Test CRUD actions on users within the django user model
    """
    
```

deployments/templates/deployment.yml
mitchdawson1982: removed

tests/users/models/test_user.py
mitchdawson1982: removed

tests/conftest.py
mitchdawson1982: The result of rebasing main which I didn't spot. Now removed

Dockerfile
mitchdawson1982: Agreed this will live outside once we utilise an external database.

MatMoore (approved):

PR #347, opened 2024-05-17
fix template so table details are shown in details pages

LavMatt:
Very small one to quickly fix the table description not being loaded to the details page 

(also removed a property that doesnt exist anymore from search results template as weirdly this errored for me locally)


hjribeiro-moj (approved):
LGTM

PR #349, opened 2024-05-17
add slack webhook secret to k8s and deploy workflows

LavMatt:


.github/workflows/deploy-generic.yml
```diff
@@ -91,6 +94,7 @@ jobs:
           NAMESPACE: ${{ secrets.KUBE_NAMESPACE }}
           ENABLE_ANALYTICS: ${{ vars.ENABLE_ANALYTICS }}
           ANALYTICS_ID: ${{ vars.ANALYTICS_ID }}
+          SLACK_WEBHOOK: ${{ secrets.slack_webhook }}
```
mitchdawson1982, line 97: perhaps define in uppercase for consistency with the other secrets.

.github/workflows/deploy.yml
```diff
@@ -19,6 +19,7 @@ jobs:
       ecr_role_to_assume: ${{ secrets.DEV_ECR_ROLE_TO_ASSUME }}
       secret_key: ${{ secrets.SECRET_KEY }}
       catalogue_token: ${{ secrets.CATALOGUE_TOKEN }}
+      slack_webhook: ${{ secrets.SLACK_ALERT_WEBHOOK }}
```
mitchdawson1982, line 22: Do we have both `slack_webhook` and `SLACK_ALERT_WEBHOOK` defined?

mitchdawson1982 (approved):
LGTM

PR #355, opened 2024-05-21
Replace all underscores with spaces in the search query

murdo-moj:
- This is a workaround for the search behaviour having underscores forcing an "OR" for all words
- This will be reverted when the search behaviour aligns again with expectations

### DIFF
--- a/home/service/search.py
+++ b/home/service/search.py
@@ -71,7 +71,7 @@ class SearchService(GenericService):
     def _get_search_results(self, page: str, items_per_page: int) -> SearchResponse:
         form_data = self.form_data
 
-        query = form_data.get("query", "")
+        query = form_data.get("query", "").replace("_", " ")
         sort = form_data.get("sort", "relevance")
         domain = form_data.get("domain", "")
         subdomain = form_data.get("subdomain", "")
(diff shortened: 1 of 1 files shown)


murdo-moj:
image

murdo-moj:
image

MatMoore:
This looks it's working on dev 👍🏻 

I tried searching for a thing with underscores, with and without quotes and it works fine.

home/service/search.py
```diff
@@ -71,7 +71,7 @@ def _build_entity_types(_, entity_types: list[str]) -> tuple[ResultType]:
     def _get_search_results(self, page: str, items_per_page: int) -> SearchResponse:
         form_data = self.form_data
 
-        query = form_data.get("query", "")
+        query = form_data.get("query", "").replace("_", " ")
```
MatMoore, line 74: Maybe add a comment here pointing at the datahub behaviour that introduced the bug? Then we can revert this if/when they fix it

MatMoore (dismissed):

home/service/search.py
murdo-moj, line 74:
```suggestion
        # Workaround for https://github.com/datahub-project/datahub/issues/10505
        query = form_data.get("query", "").replace("_", " ")
```

LavMatt (approved):

Issue #348, opened 2024-05-17
Automate the generation of a static metadata spec from Pydantic


murdo-moj:
https://github.com/ministryofjustice/find-moj-data/blob/main/lib/datahub-client/data_platform_catalogue/entities.py

- Pydantic entities represent our current metadata model
- Pydantic code is not very readable, so we want to include a page in the frontend which formats the metadata into a more readable format
- The generation of this page should probably run as part of the docker build. The script can be added to https://github.com/ministryofjustice/find-moj-data/blob/main/scripts/import-static.sh to run it as part of the npm build, or it can be another script which is explictly called somewhere in the setup of the app.
- We are trying to copy the column format of the [old metadata spec](https://docs.google.com/spreadsheets/d/1O1EjO96lyqDbyuzFWU8SNBky3b7tKM8dccZVS7-_t9g/edit#gid=1105700365). Specifically, the columns which are useful are:

Field | Type | Description | Populated by service user | Mandatory | Alterable | Example
-- | -- | -- | -- | -- | -- | --

- The pydantic code will need to be updated with a load of examples to make these fields populated.
- Depending on where the spec is hosted, we might need a new Django path

- Should version can be represented for the spec? How? Git hash?

PR #356, opened 2024-05-21
Fmd 348 metadata spec

murdo-moj:
- Added a page to display the metadata spec formed from the pydantic classes
- The page is dynamically created, but could be statically compiled as it won't change within a version.
- Missing at the moment is an indication as to which fields are filled in by the service user. There is not a concept of this in Pydantic. This can be left out of this page for now.
- The dataset registration page would make it clear which fields the user supplies. 
- Loads of examples for Pydantic classes have been added. As `Entity` subclasses have different URNs I have overloaded the urn attribute on subclasses eg the `Chart` class to display a chart urn example. 
- Something to consider: should fields with types that are classes eg `Entity.Domain` have examples or should the example be a link to the `Domain` class?
- Fixed a bug in `get_table_details` where `urn` isn't passed to `Table()`
- How can text wrapping on examples work better?

### DIFF
--- a/home/templatetags/format_metadata_field_type.py
+++ b/home/templatetags/format_metadata_field_type.py
@@ -1,0 +1,18 @@
+from django import template
+
+register = template.Library()
+
+
+@register.filter
+def format_metadata_field_type(value: dict) -> str:
+    if value.get("title"):
+        title = value["title"]
+        if value.get("type"):
+            type_string = value["type"]
+        if value.get("anyOf"):
+            type_string = " or ".join(item["type"] for item in value["anyOf"])
+        output = f"{title} ({type_string})"
+    if value.get("allOf"):
+        linked_entity = value["allOf"][0]["$ref"].split("/")[2]
+        output = f"{linked_entity} (object)"
+    return output
--- a/lib/datahub-client/data_platform_catalogue/entities.py
+++ b/lib/datahub-client/data_platform_catalogue/entities.py
@@ -17,9 +17,16 @@ class EntityRef(BaseModel):
     A reference to another entity in the metadata graph.
     """
 
-    urn: str = Field(description="The identifier of the entity being linked to.")
+    urn: str = Field(
+        description="The identifier of the entity being linked to.",
+        examples=[
+            "urn:li:chart:(justice-data,absconds)",
+            "urn:li:glossaryTerm:f045591e-182a-45fd-9b06-ebbc222606d6",
+        ],
+    )
     display_name: str = Field(
-        description="Display name that can be used for link text."
+        description="Display name that can be used for link text.",
+        examples=["absconds", "Common platform"],
     )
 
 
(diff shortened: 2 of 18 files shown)


murdo-moj:
#348

murdo-moj:
image

MatMoore:
Regarding the text wrapping, I think you can solve it with some css changes:

1. add a max-width to the column so it doesn't expand indefinitely
2. apply [overflow:auto](https://developer.mozilla.org/en-US/docs/Web/CSS/overflow) to the table so it is allowed to scroll horizontally

(Let me know if you want a hand with this)

home/templatetags/format_metadata_field_type.py
```diff
@@ -0,0 +1,18 @@
...
+
+
+@register.filter
+def format_metadata_field_type(value: dict) -> str:
+    if value.get("title"):
+        title = value["title"]
+        if value.get("type"):
+            type_string = value["type"]
+        if value.get("anyOf"):
+            type_string = "|".join(item["type"] for item in value["anyOf"])
```
MatMoore, line 13: I noticed this displays `(string|null)` for optional strings. Could we make this a bit more user friendly, e.g. `(optional string)` or `(string or null)`

lib/datahub-client/data_platform_catalogue/entities.py
```diff
@@ -28,9 +35,22 @@ class ColumnRef(BaseModel):
     A reference to a column in a table
     """
 
-    name: str = Field(description="The column name as it appears in the table")
-    display_name: str = Field(description="A user-friendly version of the name")
-    table: EntityRef = Field(description="Reference to the table the column belongs to")
+    name: str = Field(
+        description="The column name as it appears in the table",
```
MatMoore, line 39: I'm not sure this description/example are accurate any more - I think this is set to a field path so it would be `table_name.custody_id`

(It should be consistent with the `Column` class)

templates/metadata_specification.html
```diff
@@ -0,0 +1,53 @@
...
+            Classes build upon <a href="https://datahubproject.io/docs/metadata-modeling/metadata-model/#the-core-entities">
+                DataHub's metadata model.
+            </a>
+        </p>
+        {% for entity, schema_json in entities.items %}
+        <table class="govuk-table">
+            <caption class="govuk-table__caption govuk-table__caption--m" id="{{ schema_json.title }}">
+                {{schema_json.title}}
+            </caption>
+            <caption class="govuk-table__caption"><span style="font-weight:normal">{{schema_json.description}}</span></caption>
```
MatMoore, line 32: Rather than wrapping this is in a span, I think you can add one of these helper classes to the caption: https://design-system.service.gov.uk/styles/font-override-classes/#:~:text=%3C/p%3E-,Font%20weight,-As%20with%20the

It's generally a good idea to avoid inline styles in the html as it makes it hard to keep stuff consistent.

templates/metadata_specification.html
```diff
@@ -0,0 +1,53 @@
+{% extends "base/base.html" %}
+{% load markdown %}
+{% load static %}
+{% load format_metadata_field_type %}
+
+{% block content %}
+
+<div class="govuk-grid-row">
```
MatMoore, line 8: Can we add a heading to the top of the page?

MatMoore (approved):

MatMoore (approved):

MatMoore (approved):

Issue #210, opened 2024-03-22
Error screens for no connection to datahub - datahub is not responding


murdo-moj:
- We don't do error handling that well
- For example, if the search call the datahub fails, there's a visible django traceback on find-moj-data
- check for existing patterns 
- this is a custom informative error page for if the `data_platform_catalogue` library queries the GraphQL API and it has a 500 error. At the moment, we get the generic 500 error page for find-moj-data, which is uninformative as to if the frontend or backend has caused an error

MatMoore:
https://docs.djangoproject.com/en/5.0/topics/http/views/#customizing-error-views

MatMoore:
Overriding the 500 handler won't work as we can't access the original exception there:

```
    # All handlers take (request, exception) arguments except handler500
    # which takes (request).
```

Django handles this here https://github.com/django/django/blob/94ab56ee2e24d4764296580da66dbbdc9ba03b02/django/core/handlers/exception.py#L140
https://github.com/django/django/blob/94ab56ee2e24d4764296580da66dbbdc9ba03b02/django/urls/resolvers.py#L731

So looks like we need to implement this as a middleware https://docs.djangoproject.com/en/5.0/topics/http/middleware/#process-exception

PR #368, opened 2024-05-23
Catalogue unavilable page

MatMoore:
https://github.com/ministryofjustice/find-moj-data/issues/210

Add some middleware so we can show a more specific error when Datahub goes down.


mitchdawson1982 (approved):
LGTM

PR #374, opened 2024-05-24
Change sentry dsn env var to avoid datahub telemetry clash

mitchdawson1982:
Removed the existing `SENTRY_DSN` environment variable and set a new one of `SENTRY_DSN_WORKAROUND` to avoid the clash with the datahub telemetry code invoking the sentry_sdk init method with the standard variable.

### DIFF
--- a/core/settings.py
+++ b/core/settings.py
@@ -185,7 +185,7 @@ ENABLE_ANALYTICS: bool = (
 
 # Sentry Configuration
 sentry_sdk.init(
-    dsn=os.environ.get("SENTRY_DSN"),
+    dsn=os.environ.get("SENTRY_DSN_WORKAROUND"),
     enable_tracing=True,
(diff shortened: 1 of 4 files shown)


core/settings.py
```diff
@@ -185,7 +185,7 @@
 
 # Sentry Configuration
 sentry_sdk.init(
-    dsn=os.environ.get("SENTRY_DSN"),
+    dsn=os.environ.get("SENTRY_DSN_WORKAROUND"),
```
murdo-moj, line 188:
```suggestion
    dsn=os.environ.get("SENTRY_DSN_WORKAROUND"), # Datahub overwrites with this variable unless it is renamed, causing Sentry to tag issues with the incorrect environment
```

murdo-moj (approved):

Issue #351, opened 2024-05-20
Filter out CaDeT Athena entities from FMD front end


seanprivett:
### User Story
As a user of the data catalogue
I expect to not be confused/overwhelmed by duplicate entities
So i have a simple search experience

### Context
Currently `find-moj-data` treats datahub dbt models/sources/seeds as Tables and also treats Athena Datasets as tables. This means search results present what appear as duplicate entities

### Proposal
We only return the dbt entity type (be that model, seed or source) from datahub in any search queries and filter out the assocaited athena entity from all results - This can remain as a table entity for now but will likely need some input from UR around how we best classify dbt data sourced from create-a-dervied-table

### Definition of done
Agree implementation approach
Should mean the duplicate tests created [here](https://github.com/ministryofjustice/find-moj-data/pull/362) pass

LavMatt:
My current thinking for best way to implement this filter is to load in a `target_platform_instance` via the dbt ingestion recipe e.g. `target_platform_instance: cadet`. This could then be used to filter out results where platform is `athena` and platform instance is `cadet` in our search graphql query.

Unfortunately it appears from some test ingestions i've been running in dev that the native datahub dbt ingestion does no properly populate a platform instance in that it does not appear in the `dataPlatformInstance` property via a graphql query. 

I was able to properly populate a platform instance (and return the instance `cadet` from a graphql query) using the python client e.g.
```python
metadata_event = MetadataChangeProposalWrapper(
    entityUrn="urn:li:dataset:(urn:li:dataPlatform:athena,cadet.awsdatacatalog.bold_sm_spells.prison_spells_offences_dim,PROD)",
    aspect=DataPlatformInstanceClass(
        platform=mce_builder.make_data_platform_urn("dbt"),
        instance=mce_builder.make_dataplatform_instance_urn(
                    mce_builder.make_data_platform_urn("dbt"),
                    "cadet",
                )
    ),
)
client.graph.emit(metadata_event)
```
Also specifying a platform_instance in an Athena ingestion properly populates the `dataPlatformInstance` property

Need to investigate further if this is actually the intended behaviour of datahub's dbt ingestion or a bug

LavMatt:
asked question re. above to ingestion thread in datahub slack https://datahubspace.slack.com/archives/CUMUWQU66/p1716979453356899

LavMatt:
[...]

PR #383, opened 2024-05-30
add `display_in_catalogue` tag to filter in search method

LavMatt:
This PR adds the `display_in_catalogue` tag to the filter of every search meaning if that tag is not associated with an entity then that entity will not be returned in search results

Also did a bit of housekeeping and fixed some broken tests and added a workflow to run the unit tests specific to our python datahub client library. Must've been missed when the library was moved from data-platform repo

https://github.com/ministryofjustice/find-moj-data/issues/351


hjribeiro-moj (approved):
LGTM

PR #395, opened 2024-06-04
Add tags to display

LavMatt:
This PR does a couple of things:

1. It changes the tag used in all search filters from `display_in_catalogue` to `dc_display_in_catalogue`
2. It implements a new property, `tags_to_display`, in the `search_types.SearchResults` class. This is populated in a post init and makes a list useing the tags property but excludes any tag with the prefix `dc_`.


MatMoore (approved):
Looks good.. my only question is do we need to separate out `tags` and `tags_to_display`? Or can we just replace the existing `tags` with the filtered list?

LavMatt:
> Looks good.. my only question is do we need to separate out `tags` and `tags_to_display`? Or can we just replace the existing `tags` with the filtered list?

I thought worth keeping both as we may want to have other `dc_` prefixed tags we don't display to users available to use to implement some logic

PR #406, opened 2024-06-06
Add ip access restriction to the `/metrics` path only

mitchdawson1982:
Moves IP address restriction to the `/metrics` path.

### Note
IP addresses specified are the existing whilst I test, this will be reduced to only the Prometheus server / NAT ip's once clarified.


hjribeiro-moj (commented):
LGTM

hjribeiro-moj (approved):

Issue #385, opened 2024-05-31
Align dev find-moj-data domains to current cadet domains


LavMatt:
### User Story

As a developer
I expect the dev env domains to be aligned with our currently used domains
So that we can properly use dev for development

### Proposal

Following completion of this issue https://github.com/ministryofjustice/find-moj-data/issues/343 we should have entities assigned to domains as they're defined in create-a-derived-table. 

Our dev env still has an old iteration of domains that we used during early development. We should align the code to use the current domains for all environments so tidy up the code in our domain model https://github.com/ministryofjustice/find-moj-data/blob/main/home/forms/domain_model.py#L19

Otherwise searches by domain will not return anything in the dev env

We want to dynamically pull the domains from datahub rather than reference a list of hardcoded domains within the `find-moj-data` app

### Definition of Done

- [ ] All environments use the same version of domains
- [ ] dev filter by domain works as expected
- [ ] fix hardcoded domains in findmojdata so that they align with cadet/datahub automatically

PR #407, opened 2024-06-06
Populate domains drop down with what's been ingested in datahub

MatMoore:
Resolves https://github.com/ministryofjustice/find-moj-data/issues/385

Our hardcoded domain list has drifed from the CaDeT domain model. We now have a way to ingest domains from CaDeT, so we can dynamically populate our domain model in the service.

On dev, this now looks like this (there are some old domains hanging around still)

Screenshot 2024-06-10 at 12 07 01

I've sorted these alphabetically and removed the subdomains drop down for now. This is based on there being no subdomains available to select any more, regardless of which domain you choose.

Also noticed a small bug where domains weren't being displayed for Chart results, so fixed this in the process.


home/forms/domain_model.py
```diff
@@ -12,109 +16,22 @@ class DomainModel:
     Store information about domains and subdomains
     """
 
-    def __init__(self):
+    def __init__(self, search_facet_fetcher: SearchFacetFetcher):
```
MatMoore, line 19: TODO:

- This class should probably live in models, not services
- Allow this to be instantiated with values instead of depending on the service
- Change calling code to use the service instead of instantiating both this and the service

lib/datahub-client/CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## Unreleased
+
+- Return domain metadata for Charts
```
murdo-moj, line 12:
```suggestion
- Return domain metadata for Charts
- Added cache
```

MatMoore: Those changes are in the main django app, but this changelog is just for the library bit (since we are still publishing it separately)

murdo-moj:
Because of how our search is structured, only domains for which there are one of the datatypes we have defined are being pulled through. Is this the intended behaviour? eg electronic_monitoring https://datahub-catalogue-dev.apps.live.cloud-platform.service.justice.gov.uk/domain/urn:li:domain:electronic_monitoring/Entities?is_lineage_mode=false

murdo-moj (approved):

PR #416, opened 2024-06-10
Fmd 366 add dataset lineage link

LavMatt:
This PR ultimately adds to the `getDatasetDetails` query to retrieve upstream and downstream lineage for a given urn. If the lineage exists, a link to the visualisation in DataHub is rendered at the bottom of the dataset details page. 

It makes use of the existing relationships property of the base `Entity` model.

Changes include:
- 2 aliases have been added to`getDatasetDetails.graphql` - `downstream_lineage_relations` and `upstream_lineage_realtions` and these contain the respective entities for that lineage for the urn passed to the query. 
- `graphql_helpers.parse_relations()` has been refactored to parse lists of relations - so it can no handle the lineage relations returned in 2 separate blocks (those detailed above)
- Removed reference to data product parents in `search.graphql`. This was now redundant info but we could change if it's useful to return parents within search results.
-  Addition of `RelationshipType.DATA_LINEAGE` in entities
- Addition of `get_has_lineage()` method to the `DatasetDetailsService` class. This identifies if a dataset has lineage based on it's relationships returned.
- URL for lineage link added to `details_table.html` template added conditionally if has_lineage is true
- tests updated where necessary

I have met with Alex and Jess and they have given input into the design and positioning of the lineage link and label


home/service/details.py
```diff
@@ -82,13 +84,34 @@ def __init__(self, urn: str):
...
         return {
             "table": self.table_metadata,
             "parent_entity": self.parent_entity,
             "dataset_parent_type": self.dataset_parent_type,
             "h1_value": "Details",
+            "has_lineage": self.get_has_lineage(),
+            "lineage_url": f"{split_datahub_url.scheme}://{split_datahub_url.netloc}/dataset/{self.table_metadata.urn}/Lineage?is_lineage_mode=true&",
         }
 
+    def get_has_lineage(self) -> bool:
```
MatMoore, line 100: I would name this just `has_lineage` - the `get` prefix is redundant here

home/service/details.py
```diff
@@ -82,13 +84,34 @@ def __init__(self, urn: str):
...
+        """
+        Inspects the relationships property of the Table model to establish if a
+        Dataset has any lineage recorded in datahub.
+        """
+        has_lineage = (
```
MatMoore, line 105: this expression can be shortened to

`len(self.table_metadata.relationships.get(RelationshipType.DATA_LINEAGE, [])) > 0`

(the ternary is just mapping true to true and false to false)

lib/datahub-client/data_platform_catalogue/client/search.py
```diff
@@ -256,16 +255,12 @@ def _parse_result(
...
-        relations = parse_relations(
-            RelationshipType.PARENT, entity.get("relationships", {})
-        )
         domain = parse_domain(entity)
 
         metadata = {
             "owner": owner.display_name,
             "owner_email": owner.email,
             "total_parents": entity.get("relationships", {}).get("total", 0),
-            "parents": relations[RelationshipType.PARENT],
```
MatMoore, line 268: how come this is removed? Is this because we're not using it on the search results page?

MatMoore (commented):
Looks good - just a minor naming suggestion

lib/datahub-client/data_platform_catalogue/client/search.py
LavMatt: It was all geared towards data products in the search query, which i've removed. We didn't use the information anyway but i think one for another issue if we want to capture parent containers and present in the search results

MatMoore (approved):

Issue #401, opened 2024-06-05
ESDA Front end design for entries to be assigned to glossary terms


seanprivett:
## Requirements:
- [ ] Have a link to the glossary definition of an ESDA on ESDA entities
Slack thread for reference: https://mojdt.slack.com/archives/C03QZ776JVA/p1713884720156419

## Backend considerations

- Modify the GraphQL queries for search and dataset details to include glossary terms
- [This should include the glossary term description](https://datahubproject.io/docs/graphql/objects/#glossarynodeproperties)
- Return this information in the objects returned by the client library

## Frontend considerations
- On the search result page, if tagged to a glossary term of ESDA, make sure we show ESDA
- Since this only needs to work for ESDA, we can hard code the link to the ESDA glossary term URN
- Show context help with the glossary term description when ESDA is clicked
  - The LPA help component can be found here: https://github.com/ministryofjustice/opg-lpa/blob/ae5e2b9059712a98fbdbedbb315ccad3ab63f4e0/service-front/assets/js/moj/moj.modules/moj.help-system.js#L17
  - The MOJ Design System has a [Get Help](https://design-patterns.service.justice.gov.uk/patterns/get-help/) pattern which my be easier to integrate than a copy/paste of the LPA code
  - Yet another option is using `<dfn>` tags with `title` attributes. Semantically nice, but by default will just render as _italic_ and not be obvious that additional content is available, so would need some custom styling.
- Do the same on the table/database/chart details pages

See figma: https://www.figma.com/design/zTOVhHRnWfxRJqH6Gr8cIc/Catalogue-sketches?node-id=884-2316&t=hDROykODzq3zLW2C-4

## Acceptance criteria
- [ ] [All ESDAs](https://datahub-catalogue-test.apps.live.cloud-platform.service.justice.gov.uk/glossaryTerm/urn:li:glossaryTerm:83437d59-9dff-4472-bd4f-0c367cf79ac9/Related%20Entities?is_lineage_mode=false) are visibly marked as ESDA in the Find MOJ Data interface
- [ ] Whenever we show ESDA, users can interact with it to get the definition of an ESDA
- [ ] It should be clear that additional context help is available
- [ ] Context help should comply with accessibility requirements

PR #418, opened 2024-06-10
Cleanup - bring through tags and glossary terms consistently, and remove dead code for data products

MatMoore:
Part of https://github.com/ministryofjustice/find-moj-data/issues/401

This is some prep work for marking out ESDAs (via the glossary term). Frontend changes to follow.

In the process I noticed a lot of references to data products still hanging about, so I've removed those too.

### DIFF
--- a/lib/datahub-client/data_platform_catalogue/client/graphql/search.graphql
+++ b/lib/datahub-client/data_platform_catalogue/client/graphql/search.graphql
@@ -80,6 +80,28 @@ query Search(
               }
             }
           }
+          tags {
+            tags {
+              tag {
+                urn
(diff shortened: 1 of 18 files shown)


lib/datahub-client/data_platform_catalogue/client/graphql/search.graphql
```diff
@@ -179,9 +212,12 @@ query Search(
             }
           }
         }
-        ... on DataProduct {
```
MatMoore, line 182: DataProduct shouldn't be here anymore should it?

murdo-moj (approved):

PR #421, opened 2024-06-11
Show when stuff is an ESDA

MatMoore:
This only applies to a handful of assets, so in the case that a database is an ESDA, we will display this extra paragraph with a details component. This copy ideally needs to be rewritten in plain english at some point.

I've also removed metadata fields we have not populated yet (these will always display as blank) - this should be improved as part of https://github.com/ministryofjustice/find-moj-data/issues/423

We are deliberately not adding the ESDA notice on the search results page as we want this to be more concise. If we decide to show tags / glossary terms on the results page, we could potentially include ESDA there but at the moment we're not showing this.

Screenshot 2024-06-12 at 11 38 54

### DIFF
--- a/home/service/details.py
+++ b/home/service/details.py
@@ -17,6 +17,11 @@ class DatabaseDetailsService(GenericService):
             raise ObjectDoesNotExist(urn)
 
         self.result = search_results.page_results[0]
+
+        self.is_esda = any(
+            term.display_name == "ESDA" for term in self.result.glossary_terms
+        )
+
         self.entities_in_database = self._get_database_entities()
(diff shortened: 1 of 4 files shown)


alex-vonfeldmann:
so the correct spelling is MOJ not MoJ, so we'll need to fix that. Sorry my bad!

MatMoore:
> so the correct spelling is MOJ not MoJ, so we'll need to fix that. Sorry my bad!

Thanks, fixed this now

mitchdawson1982 (approved):
LGTM

PR #432, opened 2024-06-13
Add `get_database_metadata` method, add slack channel info to contact info partial for chart, dataset, and database views

tom-webber:
## Context: 
`dc_` prefixed metadata (`slack_channel` and `where_to_access`) have been added to ingestion, and we want to use `dc_slack_channel_*` to fill out contact details on the details pages.

Table (Dataset) and Chart details pages are populated via `get_[entity_type]_metadata()` methods, but the `get_database_metadata()` function is not yet implemented. This PR adds the `get_database_metadata()` method, and replaces the pre-exisitng workaround throughout the codebase, so that the implementation for populating the slack_channel in the details pages lines up with that of Table and Chart.

The `list_database_tables()` method that runs a search query to fetch database tables has been replaced with an expansion of the native `entities` fragment in the `getContainerDetails` graphql query used by `get_database_metadata`, meaning only one call is used for database details. These details are only those necessary for the database details page.

## Changes:
- Add slack channel info to `contact_info` partial for chart, dataset, and database views
 - implement `get_database_details` method
 - add `getContainerDetails.graphql` adapted from native datahub query 
   - fetch Datasets via the `entities` of the container
 - update tests
 - lots of linting, from `pre-commit` recommendations


templates/details_database.html
```diff
@@ -56,7 +56,7 @@ <h3 class="govuk-heading-s govuk-!-margin-top-3">
       </div>
 
       <div class="govuk-grid-column-one-third">
-        {% include "partial/contact_info.html" with data_owner=result.metadata.owner data_owner_email=result.metadata.owner_email %}
+        {% include "partial/contact_info.html" with data_owner=table.governance.data_owner.display_name data_owner_email=database.governance.data_owner.email slack_channel=database.custom_properties.further_information %}
```
LavMatt, line 59: should this be `data_owner=database.governance.data_owner.display_name`?

templates/partial/contact_info.html
```diff
@@ -1,20 +1,30 @@
 <div class="govuk-body">
   <h2 class="govuk-heading-s govuk-!-margin-bottom-1">IAO or Data Owner</h2>
   <p class="govuk-body">
-    {{data_owner}}<br>
-    {{data_owner_email|urlize}}
+    {{ data_owner }}<br>
+    {{ data_owner_email|urlize }}
```
LavMatt, line 5: this gives duplicate emails so suggest removing as has it's own field below

templates/partial/contact_info.html
```diff
@@ -1,20 +1,30 @@
...
   </p>
 </div>
+<!-- placeholder until we have more contact channels than just slack -->
+{% if slack_channel.url %}
+  <div class="govuk-body">
+    <h2 class="govuk-heading-s govuk-!-margin-bottom-1">Contact channels for access requests</h2>
+    <p class="govuk-body">
+        <!-- This should become a list of populated contact channels -->
+      Slack channel: <a href="{{ slack_channel.dc_slack_channel_url }}" target="_blank">{{ slack_channel.dc_slack_channel_text }}</a>
```
LavMatt, line 20: should be:
```suggestion
{% if slack_channel.dc_slack_channel_url %}
  <div class="govuk-body">
    <h2 class="govuk-heading-s govuk-!-margin-bottom-1">Contact channels for access requests</h2>
    <p class="govuk-body">
        <!-- This should become a list of populated contact channels -->
      Slack channel: <a href="{{ slack_channel.dc_slack_channel_url }}" target="_blank">{{ slack_channel.dc_slack_channel_name }}</a>

**Review by LavMatt — COMMENTED** [review]
Looking good Tom. Main question is around the graphql query and whether it needs to be as verbose.

Have a comment about the relationship for a container and whether we could improve this and remove the `list_database_tables` method but works either way

And just a couple of naming errors in the contact partial template which i've put the fixes for in the suggestion

I've not looked at the tests cos i know you and mat were doing some further work on them

File: README.md [hunk 4]
```diff
@@ -43,3 +44,7 @@ Run `pre-commit install` from inside the poetry environment to set up pre commit
 - Javascript unit tests: `npm test`
 - Selenium tests: `pytest -m tests/selenium`
 - Search benchmarks (these query the real Datahub backend): `pytest tests/benchmarks`
+
+Selenium makes use of chromedriver to run a headless browser.
+As chromedriver is updated, the local version of chromedriver and chrome may drift apart.
```
MatMoore, line 49:
```suggestion
As the chrome browser is updated, the local version of chromedriver and chrome may drift apart.
```

home/service/details.py
```diff
@@ -12,31 +12,25 @@ def __init__(self, urn: str):
...
+        self.is_esda = any(term.display_name == "Essential Shared Data Asset (ESDA)"
+            for term in self.database_metadata.glossary_terms)
         self.entities_in_database = self._get_database_entities()
         self.context = self._get_context()
 
     def _get_database_entities(self):
         # we might want to implement pagination for database children
-        # details at some point
-        database_search = self.client.list_database_tables(
+        # details    at some point
```
MatMoore, line 27:
```suggestion
        # details at some point
```

don't think this diff was intentional

lib/datahub-client/data_platform_catalogue/client/datahub_client.py
```diff
@@ -292,16 +302,59 @@ def get_chart_details(self, urn) -> Chart:
...
+                glossary_terms=glossary_terms,
                 platform=EntityRef(display_name=platform_name, urn=platform_name),
                 custom_properties=custom_properties,
             )
 
         raise EntityDoesNotExist(f"Chart with urn: {urn} does not exist")
 
-    # to expand on and replace `list_database_tables` will need new graphql query i expect
-    # but will be more equivelent to the get chart and table details methods.
+    # to expand on and replace `list_database_tables` will need new graphql query i
```
MatMoore, line 312: can this comment be removed now?

templates/partial/contact_info.html
```diff
@@ -1,20 +1,30 @@
...
+    {{ data_owner_email|urlize }}
   </p>
 </div>
+<!-- placeholder until we have more contact channels than just slack -->
+{% if slack_channel.dc_slack_channel_url %}
+  <div class="govuk-body">
+    <h2 class="govuk-heading-s govuk-!-margin-bottom-1">Contact channels for access requests</h2>
+    <p class="govuk-body">
+        <!-- This should become a list of populated contact channels -->
+      Slack channel: <a href="{{ slack_channel.dc_slack_channel_url }}" target="_blank">{{ slack_channel.dc_slack_channel_name }}</a>
```
MatMoore, line 20: [The design system generally recommends against links that open in a new tab](https://design-system.service.gov.uk/styles/links/#:~:text=link%20icon.-,Opening%20links%20in%20a%20new%20tab,-Avoid%20opening%20links)

If we keep the `target="_blank"`, we should also include "(opens in new tab)" in the link text.

tom-webber: Happy to add the extra text. Depends on expected behaviour - I would always expect a slack / teams link to open in a new window as it likely leads to migrating to a site outside of the browser.

lib/datahub-client/data_platform_catalogue/client/datahub_client.py
```diff
@@ -95,15 +97,15 @@ class DataHubCatalogueClient:
 
     If there is a problem communicating with the catalogue, methods will raise an
     instance of CatalogueError.
-    """
+    """  # noqa: E501
```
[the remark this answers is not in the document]
tom-webber: Flake8 is pretty much only calling up line length violations (those that format-on-dave with `black` doesn't auto-resolve) and unused imports (`E501`, `E401`). I think there's some use for both.

lib/datahub-client/data_platform_catalogue/entities.py
```diff
@@ -379,6 +392,7 @@ class Database(Entity):
         description="Unique identifier for the entity. Relates to Datahub's urn",
         examples=["urn:li:container:my_database"],
     )
+    database_entities: list = Field(description="list of tables in the database")
```
MatMoore, line 395: Think we can shorten this to `entities` or `tables` now, since the database part is obvious from the class.

MatMoore (approved):

MatMoore (approved):

PR #444, opened 2024-06-18
Remove phase banner

MatMoore:
The consensus is that there is no strong need for this in an internal service, so we're scrapping it.

I've moved the environment label to the header - this is still useful for the team, and is not rendered on production environments, so won't affect end users.

Screenshot 2024-06-17 at 17 02 17


mitchdawson1982 (approved):
LGTM

PR #478, opened 2024-06-21
remove actionable pagination for single page

tom-webber:
- **remove 'registered by' from search results**
- **remove page range if only single page of search results**


hjribeiro-moj (approved):
LGTM

PR #479, opened 2024-06-21
Fmd 181 display tags as links

LavMatt:
This PR ensures entities that have tags (starting non `dc_`), have those tags presented in the search and details pages as a comma separated list. It also makes these tags clickable links to a search page returning all entities with that same tag that was clicked, and that the clicked tag is present in the selected filters box.

Also adds in a new search method in the client, `get_tags()` which returns all tags in datahub. I looked at using the facet query but doesn't look like it can return more than 20 results and we already have more than 20 tags from cadet alone.

It doesn't add an independent filter component but should allow for this to be added without too much pain.

https://github.com/ministryofjustice/find-moj-data/issues/181

### DIFF
--- a/home/forms/search.py
+++ b/home/forms/search.py
@@ -119,6 +119,8 @@ class SearchForm(forms.Form):
     clear_filter = forms.BooleanField(initial=False, required=False)
     clear_label = forms.BooleanField(initial=False, required=False)
 
+    tags = forms.CharField(required=False)
+
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         self.initial["sort"] = "relevance"
--- a/home/forms/search.py
+++ b/home/forms/search.py
@@ -6,6 +6,7 @@ from django import forms
 
 from ..models.domain_model import Domain, DomainModel
 from ..service.search_facet_fetcher import SearchFacetFetcher
+from ..service.search_tag_fetcher import SearchTagFetcher
 
 
 def get_domain_choices() -> list[Domain]:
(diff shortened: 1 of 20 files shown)


home/forms/search.py
```diff
@@ -119,6 +125,8 @@ class SearchForm(forms.Form):
     clear_filter = forms.BooleanField(initial=False, required=False)
     clear_label = forms.BooleanField(initial=False, required=False)
 
+    tags = forms.MultipleChoiceField(choices=get_tags, required=False)
```
mitchdawson1982, line 128: Tags is configured as multiple choice however selecting additional tags currently overwrites the former

home/service/search_tag_fetcher.py
```diff
@@ -0,0 +1,24 @@
+from django.core.cache import cache
+
+from .base import GenericService
+
+
+class SearchTagFetcher(GenericService):
+    def __init__(self):
+        self.client = self._get_catalogue_client()
+        self.cache_key = "search_tags"
+        self.cache_timeout_seconds = 5
```
mitchdawson1982, line 10: If this is a reasonably static list could we cache it for longer ?

LavMatt: yeah we can - this just matches the other cache on domains - but they could both probably be increased

home/forms/search.py
LavMatt: I suppose it might be confusing - I think you're right and whilst the tag filter is only available via the link and always does a single choice this form choice should reflect that

LavMatt: after our discussion gonna change this to have tag filters stack up, i.e. remain multiplechoice

LavMatt: clarified in standup - the current behaviour is as expected at this stage

mitchdawson1982 (approved):
LGTM

Issue #459, opened 2024-06-18
X to remove search


seanprivett:
> could the search have an 'x' to clear action button when active and not empty? just like on gov.uk search?
> 
> 
> 
> ![Image](https://github.com/ministryofjustice/find-moj-data/assets/115145706/91a43466-bfb0-4de5-843d-f6c2c6e81477)
> 
>  

 _Originally posted by @alex-vonfeldmann in [#423](https://github.com/ministryofjustice/find-moj-data/issues/423#issuecomment-2163310598)_

YvanMOJdigital:
keyboard shortcut?

jemnery:
Also add to glossary search. Should this reset the page to all glossary terms? 🤔

MatMoore:
This behaviour actually comes from `type="search"`

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/search#differences_between_search_and_text_types

It's then customised on GOV.UK with

```css
&::-webkit-search-cancel-button {
    -webkit-appearance: none;
    background-image: url("govuk_publishing_components/icon-close.svg");
    background-position: center;
    background-repeat: no-repeat;
    cursor: pointer;
    height: 20px;
    margin-left: 0;
    margin-right: 0;
    width: 20px;
  }
}
```

We have icon-close-cross-black in the MOJ design system, so we could use that.

PR #483, opened 2024-06-24
Make search a search input, and style its close button

MatMoore:
Resolves https://github.com/ministryofjustice/find-moj-data/issues/459

This allows searches to be cancelled by clicking the cross or pressing escape.

This should be used both for the plain styled search on the glossary page, and the combined input + icon variant on the search page.

This is loosely based on the GOV.UK component: https://components.publishing.service.gov.uk/component-guide/search However, I've used the close icon already present in the MOJ design system.

In the case of the glossary, clearing the search triggers a refresh of the results. In the case of the search page, it does nothing until you search again or apply filters. We will likely revisit this soon - it would be useful to enhance this page so that the results refresh automatically without a full submit of the form.

Screenshot 2024-06-24 at 10 20 41

Screenshot 2024-06-24 at 10 17 29

### DIFF
--- a/home/forms/search.py
+++ b/home/forms/search.py
@@ -75,7 +75,9 @@ class SearchForm(forms.Form):
         max_length=100,
         strip=False,
         required=False,
-        widget=forms.TextInput(attrs={"class": "govuk-input search-input"}),
+        widget=forms.TextInput(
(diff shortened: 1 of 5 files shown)


hjribeiro-moj (approved):
LGTM

PR #488, opened 2024-06-25
DP-455 show selected filters only when filters have been applied

mitchdawson1982:
- Only displays the selected filter elements when applied filters are present.
- Removes redundant `if` statements in `selected_filters.html` partial template
- Renamed `label_clear_href` to `remove_filter_hrefs` references throughout for clarity
- Updated tests

### DIFF
--- a/templates/partial/filter.html
+++ b/templates/partial/filter.html
@@ -10,7 +10,7 @@
       </div>
     </div>
     <div class="moj-filter__content">
-      {% if label_clear_href|get_keys|length > 0 %}
+      {% if remove_filter_hrefs|get_keys|length > 0 %}
         {% include "partial/selected_filters.html" %}
       {% endif %}
       <div class="moj-filter__options">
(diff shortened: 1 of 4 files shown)


templates/partial/filter.html
```diff
@@ -10,7 +10,9 @@ <h2 class="govuk-heading-m">Filter</h2>
       </div>
     </div>
     <div class="moj-filter__content">
-      {% include "partial/selected_filters.html" %}
+      {% if label_clear_href|get_keys|length > 0 %}
```
MatMoore, line 13: [minor] Should we rename `label_clear_href` to something more descriptive?

MatMoore (approved):

templates/partial/filter.html
mitchdawson1982: I have renamed to `remove_filter_hrefs` for clarity.

MatMoore (approved):

Issue #437, opened 2024-06-17
Home page: browse by domain and search bar


seanprivett:
https://www.figma.com/design/zTOVhHRnWfxRJqH6Gr8cIc/Catalogue-sketches?node-id=703-3098&t=eAr8jTCsH7BuqGk8-0

seanprivett:
Alex and Mat working on design to present search bar on same page as domain browsing feature

alex-vonfeldmann:
very barebones design with search on home page (search stolen from gov.uk homepage)

![Image](https://github.com/ministryofjustice/find-moj-data/assets/115145706/f41a73fe-9c91-4cea-927c-fe3224897964)

MatMoore:
This one also touches the search component - maybe look at both together https://github.com/ministryofjustice/find-moj-data/issues/459

MatMoore:
The blue bit is the "masthead"

CSS here https://github.com/alphagov/govuk-design-system/blob/b3223714d8f2e9564b7e9d2ca703a3f902bf1ce8/src/stylesheets/components/_masthead.scss#L2

PR #490, opened 2024-06-25
Search and browse from the home page

MatMoore:
Ticket: https://github.com/ministryofjustice/find-moj-data/issues/437
Figma: https://www.figma.com/design/zTOVhHRnWfxRJqH6Gr8cIc/Catalogue-sketches?node-id=703-3098&t=QhJPSs4WUanjVrT8-0

I've taken the masthead component from the [design system documentation page](https://design-system.service.gov.uk/), but increased the padding a little to match the design. This uses the design system's responsive spacing, so it will take up less space on narrow screens.

There is intentionally a small margin left between the MOJ primary nav component and the masthead, because otherwise the blue underlines from the nav items bump up against the blue masthead. The design system documentation page sneakily gets around this by not including the home page in the nav, but we've decided we do want to include a "Home" link.

The variant of the search bar is based on the [GOV.UK homepage search bar](https://components.publishing.service.gov.uk/component-guide/search/homepage)

The copy within "What is domain" is a placeholder I made up for now, as the one we have in the glossary is out of date.

In addition to adding this new variant of the search box, I've updated all variants so that the button goes lighter when you hover over it (matching the component from GOV.UK publishing components).

Screenshot 2024-06-25 at 15 18 03

Screenshot 2024-06-25 at 15 18 38

### DIFF
--- a/.prettierrc
+++ b/.prettierrc
@@ -1,0 +1,3 @@
+{
+    "bracketSameLine": true
+}
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -118,9 +118,9 @@
         "filename": "templates/base/base.html",
         "hashed_secret": "f6538b22f89b1e2b05570de751f2932c6bca9969",
         "is_verified": false,
-        "line_number": 38
+        "line_number": 32
       }
     ]
   },
-  "generated_at": "2024-06-24T11:27:46Z"
+  "generated_at": "2024-06-25T08:49:19Z"
 }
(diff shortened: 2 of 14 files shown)


MatMoore:
The left / right padding of the masthead looks off on the narrow version - that's not explicitly set in the styles I took from design system docs.

Compare with https://www.gov.uk/ - there the padding is the same on all sides

hjribeiro-moj (approved):
LGTM

PR #500, opened 2024-06-26
Handle h1 values and page titles on errors

mitchdawson1982:
### Background
Existing error pages inherit from the template `base/base.html` which includes `base/head.html` utilising the `h1_value` template variable to build the page title. The `h1_value` is normally passed to render within the context however this isn't possible for error responses leading to a scenario where error pages have incomplete titles.

### Considerations
- There are multiple ways to handle custom error responses within Django, the most common [being](https://docs.djangoproject.com/en/5.0/topics/http/views/#customizing-error-views) for 400, 403, 404 and 500 errors. 
- We are already implementing custom error middleware to handle catalogue client connectivity exceptions.
- Avoid complexity by utilising one approach.

### Approach
- I have decided to implement additional checks for instances of non-catalogue-related exceptions within the `CustomErrorMiddleware.process_exception` method. We now match instances of `BadRequest`(400), and `Http404`(404) exceptions. 
- Additionally, I have chosen to implement a catch-all for any unhandled exceptions returning a `500` `Server error` response as standard. This ensures that we always render the 500 error page with the corresponding `h1_value` in all other scenarios.
- This avoids creating additional views and handling error rendering in multiple ways.


MatMoore (approved):
Nice, looks good to me 👍🏻

PR #502, opened 2024-06-26
Add a little check to make sure the pre-commit hooks ran as expected

MatMoore:
I've made this into a comment rather than a failing check, because I've noticed instances where the pre-commit hook doesn't do the right thing.

For example, djhtml refuses to process templates/navigation.html, because it contains a long line. This is because the crown logo SVG from the header contains a very long path attribute.

### DIFF
--- a/.github/workflows/check-pre-commit-has-been-run.yml
+++ b/.github/workflows/check-pre-commit-has-been-run.yml
@@ -1,0 +1,42 @@
+name: Check pre-commit hooks ran
+
+on:
+  pull_request:
+    types:
+      - opened
+      - reopened
+
+jobs:
+  check-precommit-hook:
+    runs-on: ubuntu-latest
+
+    steps:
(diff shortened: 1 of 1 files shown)


MatMoore:
It looks like the `entities` field is handled by this rolver: https://github.com/datahub-project/datahub/blob/master/datahub-graphql-core/src/main/java/com/linkedin/datahub/graphql/resolvers/container/ContainerEntitiesResolver.java

But this does not use the filter param, even though it is a valid field.

```java
    final String query = input.getQuery() != null ? input.getQuery() : "*";
    final int start = input.getStart() != null ? input.getStart() : 0;
    final int count = input.getCount() != null ? input.getCount() : 20;
```

MatMoore:
I'm closing this for now as I didn't get this working with staged changes only, and we can't run it against the whole repo while we have non-linted code in the codebase.

Issue #449, opened 2024-06-18
remove 'top level domain' label since the 'domain' header should suffice


seanprivett:
In left hand navigator, remove 'top level domain' under the Domain label

MatMoore:
I also noticed this drop down overflows its container (the filter component) when the screen is too narrow. (Haven't looked into why, but it's happening on all environments)

PR #503, opened 2024-06-26
Fmd 449 remove to level domain

LavMatt:
removes top level domain from filter side panel and fixes select box overspill by decreasing min width in base css.

#449

### DIFF
--- a/scss/base.scss
+++ b/scss/base.scss
@@ -54,3 +54,7 @@ code {
   @extend .govuk-main-wrapper;
   padding-top: govuk-spacing(3);
 }
+
+.moj-filter .govuk-select {
+  min-width: 5.5em;
+}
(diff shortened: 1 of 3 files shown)


scss/base.scss
```diff
@@ -54,3 +54,7 @@ code {
   @extend .govuk-main-wrapper;
   padding-top: govuk-spacing(3);
 }
+
+.moj-filter .govuk-select {
+  min-width: 5.5em;
```
MatMoore, line 59: Might be worth adding a comment explaining that this is decreased from the default since otherwise `moj-filter` can be narrower than this

MatMoore (dismissed):

MatMoore (approved):
👍🏻

PR #511, opened 2024-06-27
Update details page header and title

mitchdawson1982:
- Replaces static `h1_value` with names of the underlying metadata object on the details page
- Adds `<stong>` tags to denote the type of asset on the details page
- Replaces redundant bold data references in `benchmark/test_exact_matches.py` tests
- Adds a function to generate `Chart` metadata for testing purposes
- Adds a function to mock the `get_chart_details` response method


MatMoore (approved):
LGTM, but I think we need another template change to show "Chart" in a tag next to the heading, as shown here https://www.figma.com/design/zTOVhHRnWfxRJqH6Gr8cIc/Catalogue-sketches?node-id=793-3088&t=K3kBX5AXprGH8Nrf-4

MatMoore (dismissed):

Issue #447, opened 2024-06-18
Hide empty metadata items


seanprivett:
Currently done for tables, but should be implemented for all entity types on the detail page, but not on the search results page

PR #514, opened 2024-07-01
Hide empty metadata on details pages

MatMoore:
Resolves https://github.com/ministryofjustice/find-moj-data/issues/447

Currently done for tables, but should be implemented for all entity types on the detail page, but not on the search results page.

I've extracted a common base template for the details page to ensure this stays consistent. 

I've also made some minor changes that relate only to local development
- Custom error middleware should not intercept the django debug page
- Increased the long line limit to 120
- Remove azure auth from installed apps if it is disabled


LavMatt (approved):
nice

PR #515, opened 2024-07-01
Fmd 452 add feature switch - remove sort

### LINKED ISSUE #452: Remove sort results and sort radio buttons
Feature flag this so you can switch it on/off by config. This is because Scot gov are using it and may want it on.
Also allows us to do A/B testing

LavMatt:
This PR removes the sort radio button options from the top of the search results page.

It also adds in a feature switch - `search-sort-radio-buttons` that enable us to turn on and off this feature without changes to the code. 

Adds dependency on [`django-waffle`](https://github.com/jazzband/django-waffle/) - This is a package for django that allows easy definition and management of feature switches and if required more complex feature flags. Documentation here: https://waffle.readthedocs.io/en/stable/

Have added a section on feature switches to bottom of the readme

There is plenty more that can be setup and configured using waffle - e.g. feature flags with conditions so new features can be tested with specific users or sets of users and waffle endpoints that can give responses for a status of a flag or switch. To do this would require more than i've done (ie. for auth) and i'm not proposing doing yet as for the current use case the simple approach used will suffice.

resolves #452


hjribeiro-moj (approved):
LGTM

Issue #471, opened 2024-06-20
Display table name only


seanprivett:
In the search page, can we display table name instead of fully qualified name?
e.g.
current: [cadet.awsdatacatalog.bold_common_platform_linked_tables.all_offence](https://data-platform-find-moj-data-preprod.apps.live.cloud-platform.service.justice.gov.uk/details/table/urn:li:dataset:(urn:li:dataPlatform:dbt,cadet.awsdatacatalog.bold_common_platform_linked_tables.all_offence,PROD))
proposed: [all_offence](https://data-platform-find-moj-data-preprod.apps.live.cloud-platform.service.justice.gov.uk/details/table/urn:li:dataset:(urn:li:dataPlatform:dbt,cadet.awsdatacatalog.bold_common_platform_linked_tables.all_offence,PROD))

seanprivett:
Sean to create ticket on research/design board: design work required to show database parent for a table in the search results page. If we are taking the parent out of the table name display

jemnery:
Display as: **all_offence (bold_common_platform_linked_tables_all)** ??

Assumes "`cadet`" and "`awsdatacatalog`" aren't that helpful?

alex-vonfeldmann:
design for review: 

![Image](https://github.com/ministryofjustice/find-moj-data/assets/115145706/7193a3d0-10e0-4341-b47a-3fba5ffcf5c6)

PR #518, opened 2024-07-03
Fmd 471 display table name only

LavMatt:
PR addresses the long fully qualified names that were being displayed in search results for tables.
It now displays a display name for each table (which is just the table name) and the database the table is part of in the metadata below (if it has a database)

resolves https://github.com/ministryofjustice/find-moj-data/issues/471

### DIFF
--- a/lib/datahub-client/tests/client/datahub/test_search.py
+++ b/lib/datahub-client/tests/client/datahub/test_search.py
@@ -7,6 +7,7 @@ from data_platform_catalogue.client.search import SearchClient
 from data_platform_catalogue.entities import (
     AccessInformation,
     DataSummary,
+    EntityRef,
     FurtherInformation,
     TagRef,
     UsageRestrictions,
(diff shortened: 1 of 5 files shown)


templates/partial/search_result.html
```diff
@@ -7,7 +7,11 @@
       <div class="govuk-grid-column-full">
         <h3 class="govuk-heading-m govuk-!-margin-bottom-2">
           {% with result_type=result.result_type.name|lower %}
-            <a href="{% url 'home:details' result_type=result_type urn=result.urn %}" class="govuk-link">{{result.fully_qualified_name}}</a>
+            {% if result.display_name %}
+              <a href="{% url 'home:details' result_type=result_type urn=result.urn %}" class="govuk-link">{{result.display_name}}</a>
+            {% else %}
```
MatMoore, line 12: `display_name` already defaults to `name` in `parse_names` so I *think* this will always display something, and you don't need the `else`

I think I prefer having all the name defaulting logic in the library itself rather than the templates

lib/datahub-client/tests/client/datahub/test_search.py
```diff
@@ -123,6 +130,9 @@ def test_one_search_result(mock_graph, searcher):
                 metadata={
                     "owner": "",
                     "owner_email": "",
+                    "parent_container_display_name": "abc",
```
MatMoore, line 133: Could this be passed back as an `EntityRef` object rather than adding to this big dictionary? It gives us a bit more type safety and this metadata dict feels like a bit of tech debt that is no longer helping very much.

EntityRef already has both `urn` and `display_name` attributes.

LavMatt: We might need to change entity ref. If we end up having / collecting display names for databases and then their actual names we would want EntityRef to hold both. Can tackle that if it comes though i suppose

MatMoore (approved):

PR #530, opened 2024-07-11
move axe to devDependencies and not install dev deps in dockerfile

LavMatt:
relates to https://nvd.nist.gov/vuln/detail/CVE-2024-29415 - a dependency of axe, since we only use axe for selenium tests it doesn't need to be installed in our image so this PR moves it to devDependencies in package.json and add the `npm install --production` to the dockerfile which does the install without dev dependencies.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,7 +3,7 @@ FROM node:22-bullseye as node_builder
 WORKDIR /app
 COPY . .
 
-RUN npm install --production
+RUN npm install --omit=dev
 
 # The builder image, used to build the virtual environment
 FROM python:3.11-buster as builder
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -3,7 +3,7 @@ FROM node:22-bullseye as node_builder
 WORKDIR /app
 COPY . .
 
-RUN npm install
+RUN npm install --production
```
tom-webber, line 6:
```suggestion
RUN npm install --omit=dev
```
Looks like `--production` [was the old name for the flag](https://stackoverflow.com/a/9276112/10478115)

tom-webber (approved):

PR #559, opened 2024-07-16
fix reusable deploy workflow

tom-webber:
- add output to `build-and-push` job
- fix reference to job output


mitchdawson1982 (approved):
LGTM

PR #575, opened 2024-07-18
feat: add django-debug-toolbar

tom-webber:
This PR adds the [django-debug-toolbar](https://github.com/jazzband/django-debug-toolbar/tree/main) to the Django app. It adds a UI toolbar with a bunch of useful debugging tools, including current app settings / env vars, and a page-load timeline.

As currently set (default settings), this toolbar will appear during local development, as the following defaults are met:

>[`SHOW_TOOLBAR_CALLBACK`](https://django-debug-toolbar.readthedocs.io/en/latest/configuration.html#toolbar-options)
... `DEBUG` must be set to `True` and the IP of the request must be in `INTERNAL_IPS`


core/settings.py
```diff
@@ -216,6 +220,11 @@
         environment=ENV or "local",
     )
 
+# add debug toolbar when not running tests
```
MatMoore, line 223: Can we also tie this to the DEBUG setting, so that it doesn't get included in the installed apps / middleware in other environments?

tom-webber: That's the case by default:
>[SHOW_TOOLBAR_CALLBACK](https://django-debug-toolbar.readthedocs.io/en/latest/configuration.html#toolbar-options)
... used for determining whether the toolbar should show or not ... `DEBUG` must be set to True and the IP of the request must be in `INTERNAL_IPS`

core/settings.py
```diff
@@ -216,6 +220,11 @@
...
+if (not TESTING) or (not DEBUG):
```
MatMoore, line 224: I think this should be and instead of or

```suggestion
if (not TESTING) and (not DEBUG):
```

Which can be simplified to

```suggestion
if not (TESTING or DEBUG):
```

tom-webber: I've swapped these for 
```py
if DEBUG and not TESTING:
```
to take into account the above request about making sure we're reliant on `DEBUG` being set.

MatMoore (approved):

PR #576, opened 2024-07-18
Add database connection settings & psycopg library

mitchdawson1982:
With the move to new RDS instances, it was necessary to change how the database migrations and application startup are performed.

- Installs netcat on the docker container to check for postgres connectivity
- Adds psycopg library & binary for Postgres connectivity
- Updates app database settings to pull RDS values from the environment and utilises the Postgres backend when the `RDS_INSTANCE_ADDRESS` value is present. Falls back to Sqlite3 if not found.
- Creates a startup script to check for the readiness of the Postgres instance using netcat prior to migrating tables
- Startup script applies migrations
- Startup script sets up waffle switch following table migrations
- Startup script runs the app via gunicorn
- Changes the docker file entry point to utilise the startup script
- Updates the Readme file with details of how to run the app locally, utilising the RDS databases via loopback connections if required.


README.md
```diff
@@ -20,9 +23,123 @@ npm install
...
+    ```
+
+
+6) Important note for running loop back connections on Mac OS.
+Ensure that your `RDS_INSTANCE_ADDRESS` variable is set to `docker.for.mac.host.internal`,
+otherwise it will fail to connect.
+
+  ```RDS_INSTANCE_ADDRESS=docker.for.mac.host.internal```
+
+7) Change the postgres port number in ```settings.py``` and addionally in ```./scripts/app-entrypoint.sh``` if you are running as a docker image, to match the value used for the loopback connection local port value.
```
LavMatt, line 93: as in the `1234` port you've used for this command `kubectl -n data-platform-find-moj-data-dev port-forward port-forward-pod 1234:5432`

scripts/app-entrypoint.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/sh
+
+if [ -n "$RDS_INSTANCE_ADDRESS" ]; then
+    echo "Waiting for postgres..."
+
+    while ! nc -z $RDS_INSTANCE_ADDRESS 5432; do
```
LavMatt, line 6: could we make this 1234 so it aligns with the readme

core/settings.py
```diff
@@ -146,10 +146,19 @@
...
+        "ENGINE": (
+            "django.db.backends.postgresql"
+            if os.environ.get("RDS_INSTANCE_ADDRESS")
+            else "django.db.backends.sqlite3"
+        ),
+        "NAME": os.environ.get("DATABASE_NAME", BASE_DIR / "db.sqlite3"),
+        "USER": os.environ.get("DATABASE_USERNAME", ""),
+        "PASSWORD": os.environ.get("DATABASE_PASSWORD", ""),
+        "HOST": os.environ.get("RDS_INSTANCE_ADDRESS", ""),
+        "PORT": "5432",
```
LavMatt, line 158: Can this be made `1234` so aligns with the readme and works for local development without need for changing

LavMatt (commented):
Looks good Mitch - few comments re. parts of the readme that were not completely clear to me

README.md
```diff
@@ -20,9 +23,123 @@ npm install
...
 export ENV=local
 op inject --in-file .env.tpl --out-file .env
+poetry run python manage.py collectstatic --noinput
+poetry run python manage.py migrate
+poetry run python manage.py waffle_switch search-sort-radio-buttons off --create
 poetry run python manage.py runserver
 ```
 
+# Running the app against the RDS database(s)
+In order to run the app against the RDS database, you will need to do the following.
```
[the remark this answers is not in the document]
mitchdawson1982: Added addition details on this.

core/settings.py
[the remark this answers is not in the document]
mitchdawson1982: The guidance advises adding them if there is a requirement to access RDS. The values are available to the app in each Kubernetes namespace environment via a unique rds secret store.

README.md
mitchdawson1982: I have added additional clarification around this.

core/settings.py
mitchdawson1982: Not really. The value only needs to be changed when using a loopback connection and I hope the guidance is now a little clearer. I avoided creating another environment variable for the port as it will always be 5432 unless you need to target rds from your local development machine.

scripts/app-entrypoint.sh
mitchdawson1982: It is the same with the settings, this value will only need to be changed if using a loopback connection locally to target RDS.

README.md
```diff
@@ -20,9 +23,123 @@ npm install
...
+    ```
+
+5) Populate env file & add additional environment variables
+
+    ```
+    op inject --in-file .env.tpl --out-file .env
+    ```
+
+    ```
+    RDS_INSTANCE_ADDRESS=< 1pass >
```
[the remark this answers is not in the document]
mitchdawson1982: I have updated this.

README.md
```diff
@@ -20,9 +23,123 @@ npm install
 poetry run python -m nltk.downloader punkt
```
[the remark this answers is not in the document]
mitchdawson1982: I have removed this.

LavMatt (approved):

Issue #578, opened 2024-07-22
Spike: Approach for assigning ownership to users that do not exist in datahub


LavMatt:
Now we are actually assigning some real owners of cadet data to entities via our dbt cadet ingestion some issues with the approach have surfaced

As a quick recap we ask users to add a custom metadata property in cadet `dc_owner` which we then map to entity ownership in datahub. We have requested the format of the `dc_owner` value be {firstname}.{lastname} as it appears in the persons justice email so if you had `jon.smith45@justice.gov.uk` the entry would be `dc_owner: jon.smith45`. see example [here](https://github.com/moj-analytical-services/create-a-derived-table/blob/main/mojap_derived_tables/dbt_project.yml#L41)

### Issues with the current approach
The current approach does not create a user entity in datahub, although it does create an urn and assign that urn as the owner for datasets. This creates an owner against a dataset, but that owner will have no other metadata associated with it (such as email) e.g:
![Image](https://github.com/user-attachments/assets/ae034c30-9aa3-4460-8e55-990c5370dfa3)
If you were to click on the user here you would be taken to an error screen saying the entity does not exist.

This resulted in a bug where we did not display info about an owner, where we opt to show an email address (because it didn't exist) - a fix for now can be seen in this PR https://github.com/ministryofjustice/find-moj-data/pull/580

It was hoped that because we have created the urn as it would be created via an sso sign, that it would map to a user created if/when they logged into datahub, however the datahub urns appear to be case sensitive. This means you can and likely will (when cases don;t match) have duplicated users - one that is assigned to a dataset as an owner and one created at first login that will not be associated with an owned datasets (and we already do actually have this scenario). 

This is likely to cause confusion and not help if/when we want to view ownership at the user level. e.g. i want to look at this person and see all the data that they are assigned owner of.

### What should the spike achieve
Investigate and document options for approaches that would better sync ownership with user entities as they are created in datahub - without needing for every owner to log into datahub to create their user. Any option(s) should aim to avoid the issues highlighted above.

MatMoore:
Initial thoughts

- it's weird that the meta mapping does not generate an Owner entity if it is missing - find out if there is a reason for this or if we could submit a patch
- does it make any difference if we enter the whole user's email in the meta mapping? was there a reason we couldn't do that?
- we can create owners via the datahub cli
- long term, we would want a friendly ui for assigning data custodians, and it should be managed by data stewards. so this would make the meta mapping redundant anyway

[...]

PR #580, opened 2024-07-22
add _make_user_email_from_urn() to graphql helpers

### LINKED ISSUE #556: 🐛 Data owner not displayed in FMD
Data owner details are being populated at ingestion, but those are not being displayed in the front end.

LavMatt:
This PR is a fix for the owner not displaying bug. It has highlighted some deeper issues with our appraoch to ownership and usercreation.

But for a quick fix this will be ok for now so we have owners displaying in find-moj-data.

I have written up in some more dteail the issue in a new spike ticket added to our backlog, see #578 

resolves #556

### DIFF
--- a/lib/datahub-client/data_platform_catalogue/client/graphql_helpers.py
+++ b/lib/datahub-client/data_platform_catalogue/client/graphql_helpers.py
@@ -309,6 +309,6 @@ def _make_user_email_from_urn(urn) -> str:
     but for now this will fix the issue of owners being flagged in datahub but not
     showing in find-moj-data
     """
-    email_parts = urn.replace("urn:li:corpuser:", "").split(".")
-    email = f"{email_parts[0]}.{email_parts[1]}@justice.gov.uk"
+  username = urn.replace("urn:li:corpuser:", "")
+  email = f"{username}@justice.gov.uk"
     return email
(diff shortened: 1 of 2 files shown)


lib/datahub-client/data_platform_catalogue/client/graphql_helpers.py
```diff
@@ -293,3 +296,19 @@ def parse_relations(
...
+    when an urn exists for a user that has not signed into datahub via sso,
+    so has not been created as an entity and has no associated email address.
+
+    We will look to revist our approach to ownership user creation, see
+    github issue https://github.com/ministryofjustice/find-moj-data/issues/578,
+    but for now this will fix the issue of owners being flagged in datahub but not
+    showing in find-moj-data
+    """
+    email_parts = urn.replace("urn:li:corpuser:", "").split(".")
+    email = f"{email_parts[0]}.{email_parts[1]}@justice.gov.uk"
```
MatMoore, line 313: would this work? 

```suggestion
  username = urn.replace("urn:li:corpuser:", "")
  email = f"{username}@justice.gov.uk"
```

I'm not sure we should be relying on emails always having two names separated by a dot. I haven't checked in entra but seems possible that someone could have an email with 2 dots in, or none.

Could you add unit tests for this as well?

MatMoore: This will also break if we get any users that can log into entra but don't have a justice.gov.uk email. But I think it's OK as a workaround for now. Just something to consider when we pick up that other ticket.

MatMoore (approved):

PR #586, opened 2024-07-23
fmd-551: hide all tags

hjribeiro-moj:
Creates switch that manages display or hiding of result tags.

Creates a waffle switch `display-result-tags` with default set to off.

behaviour with switch set to off:

image


README.md
```diff
@@ -14,6 +14,7 @@ Please refer to Prerequisites for dependencies and installation instructions
 1. Run `poetry run python manage.py collectstatic --noinput` to collect static files
 1. Run `poetry run python manage.py migrate` this will create a local sqlite3 database and migrate any tables
 1. Run `poetry run python manage.py waffle_switch search-sort-radio-buttons off --create` to setup the waffle switch tables
+1. Run `poetry run python manage.py waffle_switch display-result-tags off --create` to setup the waffle switch tables
```
mitchdawson1982, line 17: The comment is the same as the example above

hjribeiro-moj: fixed

mitchdawson1982 (approved):
LGTM

PR #589, opened 2024-07-24
Browse by Domain Improvements

mitchdawson1982:
- Implements a new `ListDomainFetcher` class to obtain a list of domains and the number of associated entities tagged for display in the catalogue
- Only domains with entities tagged for display in the catalogue greater than zero are displayed in browse by domain on the home page
- Implements `listDomains.graphql` to list domains with query and entity filters applied
- Implements `list_domains` methods on the datahub client and search classes
- Implements a `ListDomainOption` type
- Displays the number of tagged assets associated with the given domain in the home page


LavMatt (approved):
Nice Mitch. 

I think we can probably do away with the facets too, as part of that other ticket to fix search page domain filter choices, but might be worth checking in with Mat M on Monday to double check facets code isn't doing anything else (i don't think it is)

PR #597, opened 2024-07-26
Update search form domain filters

mitchdawson1982:
This PR utilises the implementation from [here](https://github.com/ministryofjustice/find-moj-data/pull/589) to pull domains from the back end along with the count of entities tagged for display in the catalogue. 

It is key to note that this implementation bypasses the existing `facets` code, and given that we are not currently using `subdomains` and we have standalone fetchers for both tags and glossary we could consider removing the facets code altogether.

- Implements a new `get_list_domain_choices` function to return a list of `Domain` objects used for the domain filter options.

### DIFF
--- a/home/forms/search.py
+++ b/home/forms/search.py
@@ -1,10 +1,11 @@
 from copy import deepcopy
 from urllib.parse import urlencode
 
-from data_platform_catalogue.search_types import ResultType
+from data_platform_catalogue.search_types import ListDomainOption, ResultType
 from django import forms
 
 from ..models.domain_model import Domain, DomainModel
+from ..service.list_domain_fetcher import ListDomainFetcher
 from ..service.search_facet_fetcher import SearchFacetFetcher
 from ..service.search_tag_fetcher import SearchTagFetcher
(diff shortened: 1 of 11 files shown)


home/forms/search.py
```diff
@@ -19,6 +20,17 @@ def get_domain_choices() -> list[Domain]:
     return choices
 
 
+def get_list_domain_choices() -> list[Domain]:
```
MatMoore, line 23: Since nothing else is using `get_domain_choices` any more I think we can remove that implementation and replace it with this one.

I had a look at the subdomain choices and that uses a different function, `get_subdomain_choices`, so that should be unaffected. Not sure yet if we want to keep all that in the codebase but I'd prefer to wait and see until we have decided which domain model we're going to stick with.

Do you think it's worth wrapping the subdomains stuff in a feature flag so we skip any unnecessary calls?

mitchdawson1982: Awesome thanks Mat

MatMoore (approved):

PR #604, opened 2024-07-29
Linked contact href in footer to data catalogue slack

murdo-moj:
* This is in line with the contact us link in the error pages

### DIFF
--- a/templates/base/footer.html
+++ b/templates/base/footer.html
@@ -15,7 +15,7 @@
             </a>
           </li>
           <li class="govuk-footer__inline-list-item">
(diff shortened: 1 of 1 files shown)


MatMoore (approved):
Is the idea you'll switch this over once the page exists?

murdo-moj:
Do we want to direct people though the user guide for a basic slack link (2 clicks instead of 1)? I had a look at the current user guide and nothing is changed from the defaults. I assume we are going to put more time and thought into it before the prod push.

murdo-moj:
Ah yeah I see tickets to do with the user guide. I will take some.

PR #610, opened 2024-07-31
fmd-581: Add Makefile to automate running app and tests

hjribeiro-moj:
Adds Makefile with following options:

```Makefile
.PHONY: install inject_env collectstatic migrate waffle_setup run test

# Install dependencies
install_deps:
	...

# Set environment variable
set_env:
	...

# Generate .env file
generate_env:
	...

# Collect static files
collect_static:
	...

# Run migrations
migrate:
	...

# Setup waffle switches
setup_waffle_switches:
	...

# Run the application
run:
	...

# Run unit tests
test:
    ...

# Run Python unit tests
test_python:
	...

# Run JavaScript unit tests
test_js:
	...

# Clean up (optional)
clean:
	...

```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,71 @@
+# Define variables
+ENV_FILE=.env
+ENV=local
+
+# Default target
+all: build
+
+# Setup the application
+build: install_deps set_env generate_env collect_static migrate setup_waffle_switches
+
+# Install dependencies
+install_deps:
+	if ! command -v npm >/dev/null 2>&1; then \
+		echo "npm is not installed. Please install it from https://nodejs.org/"; \
+		exit 1; \
+	fi
+	if ! command -v chromedriver >/dev/null 2>&1; then \
+		echo "Chromedriver is not installed. Please install it from https://sites.google.com/a/chromium.org/chromedriver/downloads"; \
+		exit 1; \
+	fi
+	if ! command -v poetry >/dev/null 2>&1; then \
+		echo "Poetry is not installed. Please install it from https://python-poetry.org/docs/#installation"; \
+		exit 1; \
+	fi
+	poetry install
+	npm install
+
+
+
+# Generate .env file
+generate_env:
+
+	@echo "Setting ENV to ${ENV}"
+	op inject --in-file .env.tpl --out-file ${ENV_FILE}
+	@echo "Optionally, set CATALOGUE_TOKEN in ${ENV_FILE}"
+
+# Collect static files
+collect_static:
+	poetry run python manage.py collectstatic --noinput
+
+# Run migrations
+migrate:
+	poetry run python manage.py migrate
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -0,0 +1,60 @@
...
+
+# Run the application
+run:
+	poetry run python manage.py runserver
+
+# Run unit tests
+test: test_python test_js
+
+# Run Python unit tests
+test_python:
```
mitchdawson1982, line 47: You might want to consider adding the chrome-path options for the test
`poetry run pytest -vv --chromedriver-path /opt/homebrew/bin/chromedriver ./tests`

README.md
```diff
@@ -201,3 +201,56 @@ Current swicthes and default settings:
 
 - `search-sort-radio-buttons` off - switches on/off radio selection buttons for sort order of search results.
 - `display-result-tags` off - switches on/off the display of tags in search and results pages
+
+## Makefile
+
```
murdo-moj, line 206: For the readme, as a user I would just be interested in:
- `make setup`
- `make run`
- `make test`

Would `make test` replace the commands in the testing section?

MatMoore: I thinkg `make unit` `make integration` are also useful in addition to `make test`. But don't think we need to document all the build steps if we move this to the quickstart.

Don't think the makefile will completely replace manage.py, it just simplifies first time setup and hides some of the custom stuff we've bolted on, e.g. running sass

Makefile
```diff
@@ -0,0 +1,59 @@
...
+
+# Default target
+all: setup
+
+# Setup the application
+setup: install_deps set_env generate_env collect_static migrate setup_waffle_switches
+
+# Install dependencies
+install_deps:
+	poetry install
```
MatMoore, line 13: this relies on poetry and npm being installed - might want to check that along with chromedriver.

I suppose a simple way of dealing with this would just be to exit with a message that says to install these things.

Makefile
```diff
@@ -0,0 +1,59 @@
...
+	npm install
+
+
+# Generate .env file
+generate_env:
+
+	@echo "Setting ENV to ${ENV}"
+	export ENV=${ENV}
+	op inject --in-file .env.tpl --out-file ${ENV_FILE}
```
MatMoore, line 22: this is another command we assume already exists, so we should either sort this out as a dependency or give a useful error that explains how to fix it

Makefile
MatMoore: Another consideration is whether this should work on Linux as well as MacOS

We probably do at some point, as then we can use it in our gihub actions workflows, but we could come back to that later

MatMoore (approved):

README.md
hjribeiro-moj: Will create a new ticket for it

PR #633, opened 2024-08-06
Fmd 584 contact redesign

LavMatt:
PR updates the contacts partial template to the agreed improved contact design.

- The contact partial template now applies some logic within the template to render the appropriate link/text dependent on the value in the `access_requirements` custom property:
  1. a sole link given is rendered as a hyperlink with standard link text
  2. some other specific free text held in the access_requirements custom property is
  shown as given
  3. where no access_requirements custom property exists default to the standard line
- Amends some values in the new copy file and adds in new request access hyperlink text
- Adds a new function `is_access_requirements_a_url()` to details service to indicate whether the `access_requirements` custom property value is a url.
- Adds a new selenium test to make sure trhe logic is correctly rendered on the details page.
- And a unit test for the `is_access_requirements_a_url` function (and another unit test for `_parse_parent()` that wasn't added before - my bad)
- Also updated the tests workflow to compile messages as where these have changed they were not picked up by selenium tests

Kind of related as will populate the request access for all cadet models with a default but sensible link - i've done this PR but it has not been picked up https://github.com/moj-analytical-services/create-a-derived-table/pull/1883

Maybe Jake HP will have a look when he is back on 07/08 though

If you want to look at it either by deploying or running the branch locally - in the dev env i have added a test_pk table where i have put the link i've got in the above PR for the cadet request access default link

### DIFF
--- a/home/service/details.py
+++ b/home/service/details.py
@@ -10,7 +10,7 @@ from django.utils.translation import gettext as _
 from .base import GenericService
 
 
-def _parse_parent(relationships) -> EntityRef | None:
+def _parse_parent(relationships: dict) -> EntityRef | None:
     """
     returns the EntityRef of the first parent if one exists
     """
--- a/home/service/details.py
+++ b/home/service/details.py
@@ -69,7 +69,7 @@ class DatabaseDetailsService(GenericService):
             ),
             "h1_value": self.database_metadata.name,
             "is_esda": self.is_esda,
-            "is_access_requirements_url": is_access_requirements_a_url(
+            "is_access_requirements_a_url": is_access_requirements_a_url(
                 self.database_metadata.custom_properties.access_information.access_requirements
             ),
         }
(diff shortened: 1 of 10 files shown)


home/service/details.py
```diff
@@ -53,6 +69,9 @@ def _get_context(self):
             ),
             "h1_value": self.database_metadata.name,
             "is_esda": self.is_esda,
+            "is_access_requirements_url": is_access_requirements_a_url(
```
murdo-moj, line 72:
```suggestion
            "is_access_requirements_a_url": is_access_requirements_a_url(
```
potentially an improvement

MatMoore, line 72: Is the key here meant to be `is_access_requirements_a_url` to match the value?

tests/selenium/test_details_contact_contents.py
```diff
@@ -0,0 +1,79 @@
...
+from data_platform_catalogue.entities import AccessInformation, CustomEntityProperties
+
+from tests.conftest import (
+    generate_database_metadata,
+    mock_get_database_details_response,
+)
+
+
+@pytest.mark.slow
+class TestDetailsPageContactDetails:
```
MatMoore, line 11: [minor] This seems a bit find grained for a selenium test... just wondering if the unit tests will be sufficient here?

MatMoore (approved):

tests/selenium/test_details_contact_contents.py
LavMatt: I think i took to heart when Mark said the way links are rendered could be an accessibility audit fail. My thinking was this is separate logic and if something were to inadvertently change how links were rendered here it might be something we don't spot

MatMoore (approved):

PR #641, opened 2024-08-07
Hide empty matched fields on search result page

MatMoore:
DataHub was returning a spurious `tags: []` value, as part of `matchedFields` which doesn't render on the page. This meant we were showing
an empty "Matched fields" as part of the search result, even when no search query was entered.

I've also renamed our selenium test directory to integration, to make the division between unit/integration a bit clearer.

Before:
Screenshot 2024-08-07 at 15 08 47

After:
Screenshot 2024-08-07 at 16 17 16


murdo-moj (approved):
I'm behind the renaming of the tests. Otherwise the logic and tests look good.

Issue #601, opened 2024-07-29
Add text where no results


seanprivett:

tom-webber:
`Not provided`?

markjefferson-gov:
@seanprivett Couple of thoughts about this: 

* Is there any functional difference between showing some default text when the domain is blank and just assigning a domain? We have an existing domain of 'general' which is so vague I think it would suit as a catch-all bucket until we have a chance to review domains properly.
* Depending on the data, I'm assuming request access and contact channels are essential and their absence would be a dead end for the user. I appreciate this will hopefully be an edge case. How would you feel about us giving our team contact info in that scenario?

seanprivett:
@markjefferson-gov I think general is a good idea for those that don't have a domain

I think in the early days of the catalogue, entries without access or contact channel details may be the majority, so it could result in us being swamped.

markjefferson-gov:
Discussed at UCD catchup. Sean to look into likelihood of missing access and contact info.

MatMoore:
@markjefferson-gov 
> Is there any functional difference between showing some default text when the domain is blank and just assigning a domain? We have an existing domain of 'general' which is so vague I think it would suit as a catch-all bucket until we have a chance to review domains properly.

Yes - if no domain is assigned the thing will be harder to find, because it won't be reachable by browsing, and if the user interacts with the domain filter the behaviour will be a bit unintuitive.

To use general as the catchall bucket we will need to update our ingestion code for Justice Data to set the "Justice Data" dashboard to General, and also the remaining charts which don't have a suitable domain yet (I make this 52 charts)

Domains for CaDeT databases and tables are fully populated, so no change required there.

MatMoore:
Discussion regarding how to request access default:
https://mojdt.slack.com/archives/C03QZ776JVA/p1723196543249529?thread_ts=1723133391.430389&cid=C03QZ776JVA

I will add in messages to support falling back to contact or owner as an alternative to the generic message:
1. Access info is missing, but there is a contact
2. Access info and contact is missing, but there is an owner

markjefferson-gov:
Discussed and agreed required behaviour and default values with Mat and Sean on Slack. Updated the ticket description with what we agreed.

PR #656, opened 2024-08-08
Missing metadata placeholders

MatMoore:
Add some placeholder text if required metadata is missing for any reason.

Part of https://github.com/ministryofjustice/find-moj-data/issues/601

This text is provisional for now. I added two extra fallbacks for missing access info:
1. If contact info exists, direct users to that
2. otherwise, if there is a data owner, they could ask the data owner

This info *should* be mandatory in order for the asset to be displayed in the catalogue. But at the moment we have some gaps.

Screenshot 2024-08-08 at 17 02 57
Screenshot 2024-08-08 at 17 03 30
Screenshot 2024-08-08 at 17 03 40


LavMatt (approved):
LGTM

Issue #653, opened 2024-08-08
add `dc_` prefix to `access_requirements` custom property


LavMatt:
This should be done to align with other custom properties set through cadet and held within some of our pydantic model attributes

so `access_requirements` changes to `dc_access_requirements`

will need changing in:
- datahub client
- find-moj-data app
- data catalogue ingestions
- create-a-derived-table PR https://github.com/moj-analytical-services/create-a-derived-table/pull/1883

PR #663, opened 2024-08-09
update property to `dc_access_requirements`

LavMatt:
changing `access_requirements` custom property to `dc_access_requirements` so aligns with other custom properties used in cadet.

part of #653


mitchdawson1982 (approved):
🚀

PR #664, opened 2024-08-09
Stash

murdo-moj:


lib/datahub-client/data_platform_catalogue/client/search.py
```diff
@@ -28,38 +28,28 @@
...
-        )
-        self.get_tags_query = (
+        self.search_query = self.get_graphql_query("search")
+        self.facets_query = self.get_graphql_query("facets")
+        self.list_domains_query = self.get_graphql_query("listDomains")
+        self.get_glossary_terms_query = self.get_graphql_query("getGlossaryTerms")
+        self.get_tags_query = self.get_graphql_query("getTags")
+
+    @staticmethod
+    def get_graphql_query(graphql_query_file_name: str) -> str:
```
MatMoore, line 46: 🤔 I think this might be simpler as a plain function. Don't think it needs to be part of this class unless you want to treat it as a private method

lib/datahub-client/data_platform_catalogue/client/search.py
```diff
@@ -114,41 +104,61 @@ def search(
 
         logger.debug(json.dumps(response, indent=2))
 
-        page_results = self._parse_search_results(response)
+        # Should these 2 variables be bound or unbound?
```
MatMoore, line 107: Do you mean should they be instance variables? 

In general I prefer these kind of helper methods to return values rather than modifying state, otherwise you get variables that may or may not be defined depending on when you access them, which adds extra complexity

lib/datahub-client/data_platform_catalogue/client/search.py
```diff
@@ -114,41 +104,61 @@ def search(
...
+        entity = result["entity"]
+        entity_type = entity["type"]
+        entity_urn = entity["urn"]
+        matched_fields = self._get_matched_fields(result=result)
+
+        if entity_type == "DATASET":
+            try:
+                parsed_result = self._parse_dataset(entity, matched_fields, ResultType.TABLE)
+                self.page_results.append(parsed_result)
+            except Exception:
```
MatMoore, line 135: Minor thing, but I think we can have this try except surround the whole if/elif statement, rather than repeat it in each branch.

MatMoore (commented):
Nice - I think this is a really helpful change.

Have added a few minor comments but can merge after that. We should check in with @markjefferson-gov too about the error message

MatMoore (approved):

lib/datahub-client/data_platform_catalogue/client/search.py
murdo-moj: good idea!

lib/datahub-client/data_platform_catalogue/client/search.py
```diff
@@ -114,41 +94,46 @@ def search(
 
         logger.debug(json.dumps(response, indent=2))
 
-        page_results = self._parse_search_results(response)
+        # Should these 2 variables be bound or unbound?
```
murdo-moj, line 97:
```suggestion
```

PR #668, opened 2024-08-12
fix(Makefile): escape $ when launching a subshell

MatMoore:
The chromedriver variable wasn't set correctly due to Makefile quirks.

1. Each line of a makefile task runs in a separate shell, so exports need to be part of the same pipeline as the command that reads them, or be exported by Make itself via a top level `export` directive.

2. $(bla) is interpreted as a makefile variable, so we need to escape the $ for it to actually run a command.


hjribeiro-moj (approved):
LGTM

murdo-moj (approved):
hahaha I had to do this too. thanks for putting it in the makefile

PR #669, opened 2024-08-12
Feedback survey

MatMoore:
This adds a GOV.UK style survey to the page that writes to the database.

To get the data out I've added a ./manage.py command that outputs CSV.

## Screenshots

Screenshot 2024-08-13 at 15 49 42
Screenshot 2024-08-13 at 15 50 10

## Details

I've created a new app for this rather than adding to the `home` one, as they are quite separate things. The only dependency on `home` is the base layout.

I had to put a weird workaround in the ModelForm in order for the form to *not* render an empty choice with label "-----". (ModelForm behaviour is described here: https://docs.djangoproject.com/en/5.0/topics/forms/modelforms/#field-types)

I'm sure I'm missing something simple here but I was banging my ahead against it so just wanted to get something working for now.


locale/en/LC_MESSAGES/django.po
```diff
@@ -577,1576 +614,3 @@ msgstr "first name"
 #: users/models.py:12
 msgid "last name"
 msgstr "last name"
-
-#: venv/lib/python3.11/site-packages/_pytest/config/argparsing.py:475
```
MatMoore, line 581: None of this should be here... only translations from the app itself

MatMoore:
To export data:

1. Navigate to a container in k9s
2. Open a shell
3. Run `DJANGO_LOG_LEVEL=INFO ./manage.py export_feedback`

mitchdawson1982 (approved):
Nice, LGTM

PR #671, opened 2024-08-13
Remove ingress rules for the old domain

MatMoore:
The service is now on find-moj-data.service.justice.gov.uk, so everything should be using that.

The old domain is not functional anymore because the app is configured to use the new URL as the Oauth redirect_uri, meaning EntraID will redirect to the *new* origin but the app will expect cookies that were set on the *old* origin.

### DIFF
--- a/.github/workflows/notify.yml
+++ b/.github/workflows/notify.yml
@@ -50,7 +50,7 @@ jobs:
         with:
           status: ${{ github.event.workflow_run.conclusion }}
           notification_title: "Deployment Successful"
-          message_format: ":rocket: <https://data-platform-find-moj-data-preprod.apps.live.cloud-platform.service.justice.gov.uk/|New Preproduction Deployment>"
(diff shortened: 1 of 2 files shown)


.github/workflows/notify.yml
```diff
@@ -2,12 +2,13 @@ name: Notify Deployment Status
 # based on: https://github.com/integrations/slack/issues/1563#issuecomment-1588009077
 on:
   workflow_run:
-    workflows: [
-      "Deploy to Dev from branch",
-      "Deploy (with tests) to dev",
-      "Deploy to Prod on Published Release",
-      "Staged deploy to Test and Preprod"
-    ]
+    workflows:
```
MatMoore, line 5: This reformatting came from my editor - I will remove it tomorrow

hjribeiro-moj (approved):
LGTM

PR #679, opened 2024-08-15
Rename domain to subject area

### LINKED ISSUE #621: Rename domain to subject area
### What

We want to rename domain to subject area. This should be applied consistently across the service, including the home page, search results, search filter panel and data entry pages.

The home page details component (What are domains?) should be deleted.

### Why

Research revealed that users don't understand 'domain' and didn't notice or explore the details component on the home page. Our hypothesis is that subject area will be understood intuitively without need for further explanation. 

### Done when

All references to domain in the service, have been replaced with subject area.

Also update user guide to make it clear to cadet users that domains are called subject areas in our front end.

The home page details component (What are domains?) should be deleted.

MatMoore:
This removes the need for extra explanatory text on the home page.

I've also included a small change to the makefile and some documentation about reusing the code outside of MoJ, in anticipation of the meeting with Scottish Gov next week.

Resolves #621

### DIFF
--- a/.github/workflows/reusable-tests.yml
+++ b/.github/workflows/reusable-tests.yml
@@ -77,7 +77,7 @@ jobs:
         run: sudo apt-get install gettext
 
       - name: Compile messages
-        run: make compile_messages
+        run: make compilemessages
 
       - name: Setup Node.js
         uses: actions/setup-node@v4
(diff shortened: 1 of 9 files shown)


hjribeiro-moj (approved):
LGTM

PR #681, opened 2024-08-15
Make sure everything passes the linters, and enable linting via the makefile

MatMoore:
(this is not enabled yet in CI)


.pre-commit-config.yaml
```diff
@@ -56,10 +57,3 @@ repos:
     hooks:
       - id: sync_with_poetry
         args: []
-
-  - repo: https://github.com/rtts/djhtml
```
murdo-moj, line 60: does this conflict with some of the linters then? why is it removed

hjribeiro-moj (dismissed):
LGTM

.pre-commit-config.yaml
MatMoore: I ran into a couple of issues with it...

- It refuses to lint the inline SVG logo from MOJ design system, since it has an extremely long `path` attribute. It has a hard line limit of 10,000, and if you exceed that it just gives up and throws an exception, so there's no way of marking it ignored. I'd rather not fiddle with the component just to get it to work with this linter.
- It's not as optimised as some of the other linters we're using, so removing it noticeably sped up the pre-commit hook

murdo-moj (approved):

hjribeiro-moj (approved):

PR #687, opened 2024-08-20
Changed lineage description copy

murdo-moj:


mitchdawson1982 (approved):
LGTM

mitchdawson1982 (approved):
Nice

locale/en/LC_MESSAGES/django.po
```diff
@@ -627,3 +621,59 @@ msgstr "first name"
 #: users/models.py:12
 msgid "last name"
 msgstr "last name"
+
```
murdo-moj, line 624: idk what this stuff is?

MatMoore: Anything marked fuzzy is the translation system attempting to match a new msgid with an existing msgstr.

As long as you have another entry for "msgid "Enter a valid email address." etc it should be fine to remove all this.

https://www.gnu.org/software/gettext/manual/html_node/Fuzzy-Entries.html

MatMoore: Btw I'm still not sure if we wouldn't be better off using symbolic msgids instead of mapping strings to strings. The current setup seems quite confusing.

murdo-moj: those fuzzy entries are only present as fuzzy, so I will leave them for now.

Issue #590, opened 2024-07-24
Reorganise build-push-deploy workflow so build doesn't happen in environment


tom-webber:
## Context
The current separate workflows for `build-and push` and `deploy` both require workflow approvals because they run _in-environment_. `build-and-push` requires env secrets to push to the env ECR, and `deploy` deploys to the env.

## Proposal
Split `build` out of `build-and-push`, and merge `push` into `deploy` to become `push-and-deploy`. That way the build can be run once per deploy pipeline, and will not require approval.

This modification will require uploading the image artifact ([`upload-artifact`](https://github.com/actions/upload-artifact)) during the `build` workflow, then downloading the artifact ([`download-artifact`](https://github.com/actions/download-artifact)) during the push-and-deploy workflow. 

## Requirements
- [ ] Split build from build and push workflow
- [ ] Include push into deploy workflow
- [ ] Use artifact workflow-actions to pass image to the push-and-deploy workflows
- [ ] Put a sensible default for the artifact retention (can be set via repository settings)
- [ ] Rename workflows as appropriate

seanprivett:
The goal is to able to guarantee what is deployed to prod was the same build deployed to preprod. We don't want the ability to deploy direct to prod (even from main).

Proposal: creating a (draft) release deploys to preprod, awaiting approval for prod. Releases can be published once verified / deployed to prod.

At the same time, can we reduce the number of approval steps from two to one for an environment by refactoring the workflow?

PR #700, opened 2024-08-21
feat: refactor deploy pipeline to split build and deploy

MatMoore:
Part of https://github.com/ministryofjustice/find-moj-data/issues/590

The current pipeline has a couple of problems:

1. We use github's environment protection, but we access environment scoped variables in multiple jobs per environment, so we are prompted twice for approval to continue.

2. We build the docker image once per environment, so we can't guarantee the integrity of the one we end up with in prod.

To address this I have moved the ECR variables to repository level, so they can be used without setting an `environment`.

In deploy-staged.yml, we now build the image once, without requiring an extra approval step. Since we have repositories for each environment, the image is pushed to the next environment as soon as we have successfully deployed to the preceding environments.

I've also removed `deploy-to-dev.yml`, which triggered for every PR, because

1. It required us to have an otherwise unnecessary approval gate on dev, to prevent us unintentionally deploying on top of a branch somebody else is testing
2. It made it impossible to distinguish between "tests are still running" and "tests passed but the branch wasn't deployed" when looking at a list of PRs
3. We have a manually triggered `deploy-dev-from-branch.yml` which deploys a branch to dev without these problems.


MatMoore:
`ECR_ROLE_TO_ASSUME` is set by https://github.com/ministryofjustice/cloud-platform-terraform-ecr-credentials/blob/main/main.tf and configured with a list of github environments ([example](https://github.com/ministryofjustice/cloud-platform-environments/blob/main/namespaces/live.cloud-platform.service.justice.gov.uk/data-platform-find-moj-data-dev/resources/ecr.tf))

Because of this configuration, all the secrets/variables for connecting to the ECR repo are currently associated with those environments, rather than being set at the repository level.

I think we need to change this because otherwise, the `build` job needs to set the `environment` key, which we don't want. If the `environment` key is added back in then github will add the approval gate.

PR #712, opened 2024-08-22
chore: Refactor GitHub workflows for streamlined build and deploy process

tom-webber:
- Add reusable build workflow that exports image artifact
- Update deploy workflows to use new build and fetch artifact
- Rename build-and-deploy to push-and-deploy
- Remove unneeded deploy-to-dev workflow
- Integrate Slack notifications for builds


MatMoore (approved):
Nice - this is a lot simpler than my previous approach.

I think the next step will be integrating deploy-prod into the main workflow, so that we are not building the image twice.

PR #721, opened 2024-08-27
ci: move prod deploy back into staged workflow

MatMoore:
We initially thought that we should trigger prod deploys on release, as this would force us to write release notes for everything we deploy.

This idea didn't work:

1. it made it difficult to prevent skipping the earlier stages of the workflow
2. it forces every prod deploy to be a release, even if nothing user-facing has changed

This commit moves the prod release back into the deploy-staged workflow.

To help us with release notes, it automatically generates a draft release, which we can then edit and publish. I've moved this to the very end of the workflow, to give teams reusing this code the assurance that anything we publish as a release is already running in Find MoJ data prod.


murdo-moj (approved):
Give it a test on merge :)

PR #726, opened 2024-08-28
Fmd-666-tweak-error-pages

murdo-moj:
* Changes "Server error" to "There is a problem with this service" on the 500 error
* Adds a teams link to the contact us partial
* Fixed tests workflow triggers

Changing this was confusing with compiling and making messages so I've reorganised the makefile to include a compile in make messages


Dockerfile
```diff
@@ -32,7 +32,7 @@ COPY lib ./lib
 COPY locale ./
 
 RUN poetry install --without dev --no-root && rm -rf $POETRY_CACHE_DIR
-RUN make compilemessages
+RUN make messages
```
MatMoore, line 35: this should be compilemessages. The app needs the compiled versions to run

.github/workflows/reusable-tests.yml
```diff
@@ -73,11 +75,11 @@ jobs:
...
         run: poetry install --no-interaction --no-root
 
-      - name: Install compile messages prereqs
+      - name: Install messages prereqs
         run: sudo apt-get install gettext
 
-      - name: Compile messages
-        run: make compilemessages
+      - name: Make messages
+        run: make messages
```
MatMoore, line 82: this should be compiledmessages as well

Makefile
```diff
@@ -51,10 +51,7 @@ setup_waffle_switches:
 # Run makemessages
 messages:
```
MatMoore, line 52: I think it's potentially dangerous to conflate these 2 steps, since `makemessages` can generate message strings that haven't been reviewed by a human.

.github/workflows/reusable-tests.yml
```diff
@@ -1,6 +1,8 @@
 name: Run tests
 
 on:
+  pull_request:
```
MatMoore, line 4: thanks for adding this back in!!

Makefile
murdo-moj: > Then we can make sure the Dockerfile and github workflow only run compilemessages.

I have made this change, so it's all reproducible

MatMoore (approved):

PR #758, opened 2024-09-03
ci: update notifications to match new workflow structure

MatMoore:
The notifications workflow is triggered by other workflows completing, but requires the names to match. The rename should fix the prod notifications that stopped work.

Since we have now included prod in the staged workflow, this can no longer detect when preprod has finished. To workaround this, I've moved the preprod notifications directly into the staged workflow.

### DIFF
--- a/.github/workflows/notify.yml
+++ b/.github/workflows/notify.yml
@@ -2,12 +2,8 @@ name: Notify Deployment Status
 # based on: https://github.com/integrations/slack/issues/1563#issuecomment-1588009077
 on:
   workflow_run:
-    workflows: [
-      "Deploy to Dev from branch",
-      "Deploy (with tests) to dev",
-      "Deploy to Prod on Published Release",
-      "Staged deploy to Test and Preprod"
-    ]
+    workflows:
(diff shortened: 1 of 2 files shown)


.github/workflows/notify.yml
```diff
@@ -40,34 +36,10 @@ jobs:
...
-          notification-summary: "Deployment Successful"
-          notification-color: 17a2b8
-          timezone: Europe/London
-          verbose-logging: true
-
   on-prod-deploy-success:
     name: Prod Deploy Success
     runs-on: ubuntu-latest
-    if: github.event.workflow_run.conclusion == 'success' && github.event.workflow_run.name == 'Deploy to Prod on Published Release'
+    if: github.event.workflow_run.conclusion == 'success' && github.event.workflow_run.name == 'Staged deploy to Test, Preprod and Prod'
```
mitchdawson1982, line 42: I think this will send two notifications on preprod deploy success, given you have a notification block within the deploy staged workflow and this matches the same.

mitchdawson1982 (approved):
LGTM

PR #772, opened 2024-09-04
use makefile to run tests

hjribeiro-moj:
**Using makefile on test runs**

- adds new makefile targets that are to be used in githubactions (prefixed `gha_`)
- uses existing steps on tests

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -75,6 +75,8 @@ integration:
 gha_fast_tests:
 	poetry run pytest --cov -m 'not slow and not datahub' --doctest-modules
 
+gha_slow_tests:
+	poetry run pytest tests/integration --axe-version 4.9.1 --chromedriver-path /usr/local/bin/chromedriver
 
 # Clean up (optional)
 clean:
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -72,6 +88,29 @@ unit:
 integration:
 	poetry run pytest tests/integration --axe-version 4.9.1 --chromedriver-path $$(which chromedriver)
 
+
+# Install project dependencies for GitHub Actions
+gha_install_project: install_python_deps install_npm_deps
```
MatMoore, line 93: I suggest we remove these extra tasks that only run on github actions - I don't think they are gaining us anything.

I can't remember what we discussed in standup, but I understood the point of this ticket to be about removing duplication between local env setup and CI setup. If it's not easy to do that I think it's fine to leave stuff as it is.

Makefile
```diff
@@ -26,8 +42,8 @@ install_deps:
 		echo "Poetry is not installed. Please install it from https://python-poetry.org/docs/#installation"; \
 		exit 1; \
 	fi
-	poetry install
-	npm install
+	@make install_python_deps
```
MatMoore, line 45: is there a reason why these are run via `@make` rather than being a dependency of the task? Just looks a bit strange like this

Also I think you can drop the checks for npm and poetry above as they're included twice now.

MatMoore (approved):

PR #773, opened 2024-09-04
Centralised server cache

mitchdawson1982:
- Adds helper function to generate valid Cache configuration for respective environments.
- Helper function provides configuration for Elastic Cache Redis connectivity via Cloud Platform in non local development scenarios.
- Helper function provides Memcache configuration for local development

### DIFF
--- a/core/helpers.py
+++ b/core/helpers.py
@@ -25,7 +25,7 @@ def generate_cache_configuration() -> dict[str, Any]:
         )
 
         domain: str = ".".join(
-            os.environ.get("REDIS_PRIMARY_ENDPOINT_ADDRESS", "").split(".")[2:]
+            os.environ.get("REDIS_PRIMARY_ENDPOINT_ADDRESS", "").split(".")[1:]
         )  # 'iwfvzo.euw2.cache.amazonaws.com'
 
         hosts: list[str] = json.loads(
@@ -36,7 +36,6 @@ def generate_cache_configuration() -> dict[str, Any]:
             location.append(
                 f"rediss://:{os.environ.get('REDIS_AUTH_TOKEN')}@{host}.{domain}/{REDIS_DB_VALUE}"  # noqa: E501
             )
-        print(location)
 
         cache["LOCATION"] = location
 
--- a/core/helpers.py
+++ b/core/helpers.py
@@ -1,3 +1,4 @@
+import json
 import os
 from typing import Any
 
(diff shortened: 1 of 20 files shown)


core/settings.py
```diff
@@ -197,7 +197,8 @@
 
 CACHES = {
     "default": {
-        "BACKEND": "django.core.cache.backends.locmem.LocMemCache",
+        "BACKEND": "django.core.cache.backends.redis.RedisCache",
+        "LOCATION": f"rediss://:{os.environ.get('REDIS_AUTH_TOKEN')}@{os.environ.get('REDIS_PRIMARY_ENDPOINT_ADDRESS')}/0",  # noqa: E501
```
MatMoore, line 201: is the double s here a typo?

Also, what do you want to happen if REDIS_AUTH_TOKEN / REDIS_PRIMARY_ENDPOINT_ADDRESS are not set? At the moment it will construct an invalid connection string. Can we keep the LocMemCache as a fall back for local dev?

home/service/domain_fetcher.py
```diff
@@ -24,10 +24,12 @@ def fetch(self) -> list[DomainOption]:
         """
         result = cache.get(self.cache_key)
         if not result:
+            print("not found in cache")
```
MatMoore, line 27: some print lines here that can be removed

core/settings.py
mitchdawson1982: `rediss://` indicates a secure connection with tls rather than a standard one. Yes, this is just an intial change to test the functionality within the dev environment. I will add logic to fall back to memcache if local development or variables are not present.

deployments/templates/deployment.yml
```diff
@@ -99,6 +99,26 @@ spec:
...
+              valueFrom:
+                secretKeyRef:
+                  name: fmd-redis-secrets
+                  key: primary_endpoint_address
+            - name: REDIS_AUTH_TOKEN
+              valueFrom:
+                secretKeyRef:
+                  name: fmd-redis-secrets
+                  key: auth_token
+            - name: REDIS_MEMBER_CLUSTERS
```
MatMoore, line 112: are these last 2 variables needed?

MatMoore (approved):

core/helpers.py
```diff
@@ -0,0 +1,43 @@
...
+
+        hosts: list[str] = json.loads(
+            os.environ.get("REDIS_MEMBER_CLUSTERS", "")
+        )  # ["cp-f05ff2dca7d81952-001","cp-f05ff2dca7d81952-002"]
+
+        for host in hosts:
+            location.append(
+                f"rediss://:{os.environ.get('REDIS_AUTH_TOKEN')}@{host}.{domain}/{REDIS_DB_VALUE}"  # noqa: E501
+            )
+        print(location)
```
murdo-moj, line 39: Still want the print statement?

murdo-moj (approved):

Issue #674, opened 2024-08-14
add test for presence of database in table detail breadcrumbs


LavMatt:
Currently we don't have a test to check for the presence of the database in the table detail bread crumb or other details (like charts) - we should add tests for this to make sure changes don't impact this without us realising

PR #786, opened 2024-09-05
Add an end to end test that checks breadcrumbs

MatMoore:
Currently we don't have any tests that run end to end using a real Datahub backend.

This PR adds one to test that breadcrumbs are rendered correctly. https://github.com/ministryofjustice/find-moj-data/issues/674

Note: this introduces a CI dependency on the **dev datahub** instance, so it's possible that we could break that environment and block CI. However, CI is independent of the **dev Find MoJ data** instance.

As there is no test isolation, the test makes some assumptions about what data is populated:

- there is a subject area for Bold
- there is at least one table in it (from CaDeT)
- the CaDeT ingestion ran as expected, and populated database entities


.github/workflows/reusable-tests.yml
```diff
@@ -115,6 +115,13 @@ jobs:
         if: steps.fast-tests.outcome == 'success'
         run: TESTING=True poetry run pytest tests/integration --axe-version 4.9.1 --chromedriver-path /usr/local/bin/chromedriver
 
+      - name: Run integration tests
+        if: steps.fast-tests.outcome == 'success'
+        run: TESTING=True poetry run pytest tests/end_to_end --chromedriver-path /usr/local/bin/chromedriver
```
murdo-moj, line 120: You could use the makefile here?

murdo-moj (dismissed):

murdo-moj (approved):

Issue #760, opened 2024-09-03
Re-introduce IP restrictions on DataHub


jemnery:

MatMoore:
https://developers.google.com/search/docs/crawling-indexing/block-indexing
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/meta/name#other_metadata_names

seems like we should go with
`<meta name="robots" content="noindex, nofollow" />`

tom-webber:
As a note, robots.txt is a guideline rather than an enforceable rule. Some companies do respect it, but others don't. If we want to _block_ crawlers, we'll need to block their published crawler IPs.

PR #795, opened 2024-09-09
Block search engines and remove redundant <main> element from error pages

MatMoore:
- This is an internal service so we don't want it in google (https://github.com/ministryofjustice/find-moj-data/issues/760)
- The main element and govuk-width-container div were being duplicated between the base layout and the error pages


murdo-moj (approved):
Nice!

PR #802, opened 2024-09-09
Dependabot issue creation

mitchdawson1982:


.github/workflows/dependabot-create-issue.yml
```diff
@@ -0,0 +1,31 @@
...
+            if: github.event.pull_request.user.login  == 'dependabot[bot]'
+            env:
+              pr_title: ${{github.event.pull_request.title}}
+              pr_number: ${{github.event.pull_request.number}}
+              pr_url: ${{github.event.pull_request.url}}
+            run: |
+              project="Data Catalogue"
+              label="Dependabot Issue"
+              title="Dependabot PR $pr_title opened"
+              body="Dependabot has opened PR #$pr_number url - $pr_url
```
MatMoore, line 26: I think github might collapse #pr_number and #pr_url to the same thing, so you might not need both?

.github/workflows/dependabot-create-issue.yml
```diff
@@ -0,0 +1,31 @@
...
+          - name: Open issue if Dependabot PR
+            if: github.event.pull_request.user.login  == 'dependabot[bot]'
+            env:
+              pr_title: ${{github.event.pull_request.title}}
+              pr_number: ${{github.event.pull_request.number}}
+              pr_url: ${{github.event.pull_request.url}}
+            run: |
+              project="Data Catalogue"
+              label="Dependabot Issue"
+              title="Dependabot PR $pr_title opened"
```
MatMoore, line 25: Can we name this something like "Review dependabot PR: $pr_title" so it reflects the task that needs doing?

Also, where's the emoji? 🧌

MatMoore (approved):
:dependabot: 👍🏻

.github/workflows/dependabot-create-issue.yml
mitchdawson1982: Looking it [this](https://docs.github.com/en/webhooks/webhook-events-and-payloads#pull_request) they still seem to be different things -

MatMoore: I mean when it renders the text, I can type `#802` or `https://github.com/ministryofjustice/find-moj-data/pull/802` and they both get rendered as the same link, #802

.github/workflows/dependabot-create-issue.yml
mitchdawson1982: Updated and Emoji Added :-)

MatMoore (approved):

PR #822, opened 2024-09-11
ci: completely remove e2e test from ci

MatMoore:
including a test that depends on our datahub dev environment (and therefore needs a secret token) breaks in a bunch of subtle ways:

- workflow_call triggered jobs can't access the secret, unless we explicitly pass it through from the parent workflows
- dependabot can't access the secret, unless we copy the secret to dependabot secrets
- pull requests from outside the org can't access the secret either (a security measure to prevent exfiltrating secrets)

the benefit of this test is already quite minimal so this has convinced me we don't need it in our CI pipeline.

### DIFF
--- a/.github/workflows/reusable-tests.yml
+++ b/.github/workflows/reusable-tests.yml
@@ -115,15 +115,6 @@ jobs:
         if: steps.fast-tests.outcome == 'success'
         run: make integration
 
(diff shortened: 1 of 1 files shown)


mitchdawson1982 (approved):
🚀

Issue #784, opened 2024-09-05
🐛 Remove filter heading is not correctly cased


LavMatt:
### Describe the bug.

When selected for and entity type or subject area the heading of the remove label is cased incorrectly, e.g.

`Entity Type` and `Subject Area` rather than `Entity type` and `Subject area`

PR #829, opened 2024-09-12
Fix sentence casing on subject area, entity type labels when filters are selected

MatMoore:
Resolves https://github.com/ministryofjustice/find-moj-data/issues/784
and also fixes an unrelated issue with the tests, caused by the screenshot helper being moved to the top level. This massively slowed down the tests because it has autouse=True and it depends on the selenium driver being initialised.


mitchdawson1982 (approved):
Nice

PR #839, opened 2024-09-16
Quoted search fix

mitchdawson1982:
Adds a method to the search service to handle updated search behaviour in datahub. The recent datahub changes now mean that query strings containing either quotes, underscores or full stops are treated as exact matches. 

This PR ensures that quoted queries are passed through to search unchanged whereas unquoted queries containing underscores are altered to remove them and replace them with spaces, to ensure a non-exact match search is performed.


home/service/search.py
```diff
@@ -56,11 +56,17 @@ def _build_entity_types(self, entity_types: list[str]) -> tuple[ResultType, ...]
         )
         return chosen_entities if chosen_entities else default_entities
 
+    def _format_query_value(self, query: str) -> str:
+        query_pattern: str = r"^[\"'].+[\"']$"
+        # We expect a quoted query to be an exact match therefore when not, we replace
```
MatMoore, line 61: I find this comment confusing out of context 😕 

Can we say something like "datahub treats any query with underscores as exact, so if the string is not quoted, we convert underscores to spaces so that we get partial matches as well"

MatMoore (approved):
This looks good, but does it solve the bug on the ticket where "Common Platform" doesn't match anything?

https://find-moj-data.service.justice.gov.uk/search?query=%22Common+Platform%22&domain=

Even with exact matching I would expect this to return results, since the whole phrase is in the description of https://find-moj-data.service.justice.gov.uk/details/table/urn:li:dataset:(urn:li:dataPlatform:dbt,cadet.awsdatacatalog.common_platform_derived.cases_fct,PROD)

MatMoore (approved):

Issue #798, opened 2024-09-09
🐛 when browsing by subject area sort alphabetically


YvanMOJdigital:
### Describe the bug.

🐛 when browsing by subject area sort alphabetically

YvanMOJdigital:
talk about with tochi

PR #854, opened 2024-09-19
feat: sort non keyword search results a-z

LavMatt:
This PR addresses #798 and now search results when browsing by subject area or filtering will be sorted ascending rather than the default relevant ordering.

or in other words any results shown that are not the result of a keyword search query entered via the search input text box

I've also had to change the field name passed to the graphql `searchAcrossEntities` query's `sort` arg. If `name` is used it errors if no results are returned, but using `_entityName` is handled by the query and is consistent with how the datahub UI uses the query to sort search results

### DIFF
--- a/home/service/search.py
+++ b/home/service/search.py
@@ -67,7 +67,15 @@ class SearchService(GenericService):
     def _get_search_results(self, page: str, items_per_page: int) -> SearchResponse:
         form_data = self.form_data
         query = self._format_query_value(form_data.get("query", ""))
-        sort = form_data.get("sort", "relevance")
+
+        # we want to sort results ascending when a user is browsing data via non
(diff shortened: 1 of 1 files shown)


murdo-moj:
A lot of this code looks left over from when we had more customisability on the sorting of results. Is there a bit of a tidy up of that code we can do since the sort option on the form doesn't do anything now?

LavMatt:
sort is still on a switch should we want to turn it on

> A lot of this code looks left over from when we had more customisability on the sorting of results. Is there a bit of a tidy up of that code we can do since the sort option on the form doesn't do anything now?

I think we retain that because we have a waffle switch to enable us to turn sort radio buttons back on

murdo-moj (approved):

PR #872, opened 2024-09-24
docs: update request access link text

### LINKED ISSUE #871: Accessibility - change request access link text to not use click
During our meeting to run through accessibility it was raised that use of the word click is not inclusive from an accessibility perspective as not all users will be clicking when using the service.

The wording should be changed from `Click link for access information (opens in new tab)` to `Select this link for access information (opens in new tab)`

LavMatt:
resolves #871


mitchdawson1982 (approved):
LGTM

PR #896, opened 2024-09-30
Fmd 880 add content to empty table cells

LavMatt:


templates/details_dashboard.html
```diff
@@ -20,7 +20,11 @@
                 <tr class="govuk-table__row">
                   <td class="govuk-table__cell"><a href="{% url 'home:details' result_type=chart_type urn=chart.entity_ref.urn %}" class="govuk-link">{{chart.entity_ref.display_name}}</a></td>
                   <td class="govuk-table__cell">
-                      {{ chart.description|markdown:3 }}
+                  {% if chart.description %}
+                    {{ chart.description|markdown:3 }}
+                  {% else %}
+                    <p class="govuk-visually-hidden">Description for {{chart.entity_ref.display_name}} does not exist</p>
```
MatMoore, line 26: Minor nitpick but wondering if this should be "A description for ... does not exist"

MatMoore (dismissed):

templates/details_dashboard.html
LavMatt: yeah probably reads a bit better

MatMoore (approved):

PR #927, opened 2024-10-08
feat: add switch for table details is_nullable column

LavMatt:
we are turning this off because currently is nullable metadata are not accurate

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -54,6 +54,7 @@ migrate:
 setup_waffle_switches:
 	poetry run python manage.py waffle_switch search-sort-radio-buttons off --create
 	poetry run python manage.py waffle_switch display-result-tags off --create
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -54,6 +54,7 @@ migrate:
 setup_waffle_switches:
 	poetry run python manage.py waffle_switch search-sort-radio-buttons off --create
 	poetry run python manage.py waffle_switch display-result-tags off --create
+	poetry run python manage.py waffle_switch show_is_nullable_in_table_details_column off --create
```
MatMoore, line 57: Can we default it to `on` so it doesn't impact other deployments of this codebase (e.g. Scottish Gov)

We could switch it off manually via the django console

MatMoore (approved):

PR #934, opened 2024-10-10
🚧 Responsive tables

mitchdawson1982:
- Makes tables on results details pages responsive when loaded in small resolutions via custom css
- Utilises `truncatewords_html` to truncate descriptions to 25 words on linked items that can be clicked on to display the full description. This was because keeping the amount of text in the description fields relatively consistent improved the look and feel in the table view.   `{{ table.description|markdown:3|truncatewords_html:25 }}`
- I have removed the truncation of the description for tables as there is no other way to view the full description in this case.
- Updates tests


MatMoore:
A couple of thoughts after coming back to this:

1. In the block view, I think it might be useful for the column names to be followed by colons, just to make the names/values more distinct. That'll also make them consistent with how we've done the metadata items at the top of the page.
2. We're currently switching to the block style at the tablet breakpoint. I think this makes sense (and if in doubt, I would go with what the NHS have done), but it's possible the normal table style would still work at that breakpoint and we could switch it at the mobile breakpoint instead. What do you think?

I think we should also run this by the accessibility team, since they pointed us towards this component originally.

MatMoore:
For reference here is the ONS version which looks like it triggers at the "small" breakpoint (500px)

https://service-manual.ons.gov.uk/design-system/components/table/#example.example-table-responsive.html:~:text=Example%20Table%20Numeric-,Responsive%20table,-Setting%20%22variants%22%3A%20%27responsive
https://github.com/ONSdigital/design-system/blob/main/src/components/table/_table.scss#L65

murdo-moj (approved):

Issue #913, opened 2024-10-02
Implement the ingestion of the Performance hub metadata


teeceeas:

MatMoore:
The spike ticket says the result will likely look like a chart but I'll need to check whether we can actually link to the chart like we did with with Justice Data.

Both charts and datasets can be specialised with SubType. Setting this to a value like "PerformanceMetric" preserves the option of handling metrics differently to other kinds of charts.

jemnery:
This is what the ownership logic was - we're not implementing this

- If a metric has been assigned to the **probation** subject area, add "**hubprobation@justice.gov.uk**" as the owner
- For **prison**  subject area metrics:
  - If `DataSource` is "Finds Upload", "Incidents Upload" "Incidents Upload (IRS)" or "Hub upload (assaults & referrals)" assign **HubNationalOperation@justice.gov.uk** as the owner
  - If `DataSource` is "Alere Tox lab data", assign **HubMDT@justice.gov.uk** as the owner ("MDT" is mandatory drug testing in prisons)
  - If `DataSource` is "Programmes return" or "HII programme uploads" OR the `MetricRef` starts with "CUP", assign **HubProgrammesMinorities@justice.gov.uk** as the owner
  - Otherwise assign **hubprisons@justice.gov.uk** as the owner

MatMoore:
Work in progress, but have something ingested
https://dev.find-moj-data.service.justice.gov.uk/details/dashboard/urn:li:dashboard:(performance-hub,performance-hub)

We're going to need to adjust the templates in Find MoJ data

- Charts have hardcoded links to Justice data - this should be based on platform
- There are two many metrics to display on one page - we really need some kind of pagination or search
- Subject area is missing for dashboard

MatMoore:
Chart/dashboard is not actually a very good description of what these metrics are, so I'm going to change it to Dataset/Container, and introduce a subtype for Metrics.

We can then hide the schema part from the template.

PR #943, opened 2024-10-14
Hide schema section for performance hub metrics and display caveats

MatMoore:
https://github.com/ministryofjustice/find-moj-data/issues/913

Add a template specifically for metrics which hides the schema section, and adds some caveats about use of the data.

I've deliberately kept this out of the description field, because it is common to all datasets from this platform, and clutters up the details page of the Performance Hub itself (modelled as a database).

### DIFF
--- a/home/service/details.py
+++ b/home/service/details.py
@@ -94,6 +94,8 @@ class DatasetDetailsService(GenericService):
 
         self.context = self._get_context()
 
+        self.template = self._get_template()
+
     def _get_context(self):
         split_datahub_url = urlsplit(
             os.getenv("CATALOGUE_URL", "https://test-catalogue.gov.uk")
@@ -111,6 +113,13 @@ class DatasetDetailsService(GenericService):
                 self.table_metadata.custom_properties.access_information.dc_access_requirements
             ),
         }
+
+    def _get_template(self):
+        return (
+            "details_metric.html"
+            if "Metric" in self.table_metadata.subtypes
+            else "details_table.html"
+        )
 
     def has_lineage(self) -> bool:
         """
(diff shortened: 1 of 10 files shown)


LavMatt:
Looking good, have a few points/thoughts:
- I'm wondering if we should look to make use of the entity sub-type which is already returned by the client's `get_table_details()` method. We could then distinguish a metric from a table (although the sub-type for these from cadet might be model or even view so we might need some kind of mapping step too). 
- The search method also returns the sub type for each entity so we could display more precise entity types on both details and search results pages.
- Maybe not one for now but would we want to be able to filter metrics as a separate entity type on the search page?
- Just an observation on content but often the description is just repetition of the metric name and in some cases it's not obvious what the measure is from the name and description. e.g. `1 to 1 completions` - i'm not sure what this is measuring.

MatMoore:
@LavMatt I started out making use of the sub type as you describe, but I decided against it at this stage because it turned into quite a big change, and I think it's quite likely that we would end up changing or reverting any changes we make now once we actually get some feedback on it.

The areas where the entity/result type gets surfaced in Find MoJ data are:
- On the details page
- In the URL of the details page
- On search results
- In the search filter

Making the search filter distinguish between Table and Metric requires would require changes to the search function to take into account subtype, and the label on the search results page requires a change to the query and result parsing code.

I also had some doubts whether "Metric" is the right way of describing this data. It may cause some confusion because we've ingested Justice Data measures as "Charts" even though they are pretty similar to Performance Hub Metrics. So I think it's better to wait until we've spoken to generalist users before changing the data model.

LavMatt (approved):

Issue #976, opened 2024-10-23
Generalise "How to contact us" mechanism


MatMoore:
### User Story

As a team who maintains data
We want to be able to be able to choose how we receive questions about the data
(which may include multiple channels) 
So that we don't have to change how we work when our data is put on the catalogue

### Useful Contacts

@jemnery @MatMoore

### Proposal

In addition to slack, provide custom properties for 
 
- MS Teams
- A shared email inbox

Modify templates/partial/contact_info.html to display these channels if available.

### Additional Information

The shared inbox was requested by the CJS Dashboard team, who already monitor a shared inbox, and would prefer questions go there rather than their team slack.

As it stands, we only support showing the emails of a data custodian. This doesn't work for this team because
 
- it's a shared inbox, not a named individual
- its on the digital.justice.gov.uk, not justice.gov.uk so you can't log in with it

We'd always intended to add Teams in addition to Slack, but we skipped it for alpha.

PR #992, opened 2024-10-28
feat: add more contact options for ask a question 

LavMatt:
PR for #976 

It makes possible surfacing different contact information for the Ask a question section in fmd. Two additional types of ways to ask a question have been added.

- Microsoft Team channel
- Team email (not not be confused with ms teams)

I have added three additional properties to our entity class in datahub library `Entity.custom_properties.further_information` where the `FurtherInformation` class now has the properties:
- dc_teams_channel_name
- dc_teams_channel_url
- dc_team_email

For teams looks like 
image

For team email looks like
image

Where there are all three of the options present i opted not to display all of the contact info as this might be a bit confusing and make the screen very busy. I'm assuming most teams will have a preferred way to be contacted re questions about the data and will enter this for their contact property.

In the case all contact info is present the order of importance is Slack > Teams > team email

for balance here's what it would look like if we displayed everything we held (and we held all routes)
image


murdo-moj:
With multiple contact channels potentially present, I changed the options to be a list. If no channels are provided, the standard "Not provided" text is shown

murdo-moj:
image

murdo-moj:
* Slack, teams and a team email all get listed as contact channels. 
* If none of those channels exist but there's an owner, direct the user to the owner
* If no owner exists, the standard "Not provided." text is shown

hjribeiro-moj (approved):
LGTM

PR #993, opened 2024-10-28
Show data custodian if set in datahub

MatMoore:
This PR extends the `Governance` object to include data owner, data stewards, and data custodians, based on the `ownershipType` in Datahub.

If an owner of type data custodian exists, then Find MoJ data will render "Data custodian (technical owner)" rather than "Data owner".

To ease deployment, I've kept it rendering "Data owner" if the data custodian is not set. We can remove this once the ingestion has been updated to use data custodian instead of data owner.

### DIFF
--- a/lib/datahub-client/data_platform_catalogue/client/graphql_helpers.py
+++ b/lib/datahub-client/data_platform_catalogue/client/graphql_helpers.py
@@ -39,7 +39,7 @@ def get_graphql_query(graphql_query_file_name: str) -> str:
     return query_text
 
 
-def _parse_owner(owner: dict):
+def _parse_owner_object(owner: dict):
     properties = owner.get("properties") or {}
     display_name = (
         properties.get("displayName")
@@ -56,9 +56,8 @@ def _parse_owner(owner: dict):
     )
 
 
-def parse_owner(
+def parse_data_owner(
     entity: dict[str, Any],
-    ownership_type_urn: str = DATA_OWNER,
 ) -> OwnerRef:
     """
     Parse ownership information, if it is set, and return the first owner of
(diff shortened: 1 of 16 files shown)


lib/datahub-client/data_platform_catalogue/client/search.py
```diff
@@ -136,7 +137,7 @@ def _parse_search_results(self, response) -> Tuple[list, list]:
                 else:
                     raise Exception
             except Exception:
-                logger.warn(f"Parsing for result {entity_urn} failed")
+                logger.exception(f"Parsing for result {entity_urn} failed")
```
murdo-moj, line 140: do exceptions here get raised and break the ingestion?

MatMoore: No, this is the same as the `logger.error` but it will include the stack trace in the log which is quite useful

lib/datahub-client/data_platform_catalogue/client/graphql_helpers.py
```diff
@@ -32,32 +39,78 @@ def get_graphql_query(graphql_query_file_name: str) -> str:
...
+        display_name=display_name or "",
+        email=properties.get(
+            "email",
+            _make_user_email_from_urn(owner.get("urn")),
+        ),
+        urn=owner.get("urn", ""),
+    )
+
+
+def parse_owner(
```
murdo-moj, line 59: We now have `parse_owner`, `_parse_owner`, and `parse_owners`, can we make do with just the plural? This might be satisfied by a function name change

MatMoore: `parse_owner` is still used for the data owner, which I've left in the data model for now. The internal function I can rename though

MatMoore: Maybe I can replace `parse_owner` with `parse_data_owner` to avoid confusion

murdo-moj: Reading the rest of the code here's a more thought out solution:
`parse_owner` -> `parse_data_owner`
`_parse_owner` -> `parse_generic_owner`

murdo-moj (approved):

Issue #995, opened 2024-10-28
✨ Replace create issue user email field with checkbox


mitchdawson1982:
### Describe the feature request.

We can simplify the create issue form and save the user entering an email address by simply having a checkbox that signifies if the user wishes to be contacted. Additionally, instead of populating the user's email address onto the `Issue` instance, we should link the user and issue models by foreign key.

### Approach
- Add a foreign key linking the issue model with the user model
- Add `email_user` boolean field to the issue model and form
- Remove the `user_email` field from the issue model and form
- Add `email_user` checkbox to the create issue template
- Update the view to obtain the user model and link it to the issue prior to save

PR #1009, opened 2024-10-31
Don't ask for email when reporting issues

MatMoore:
https://github.com/ministryofjustice/find-moj-data/issues/995

Instead of requiring the reporter to enter their email, we can just use the email of the currently signed in user.

Instead, just ask if they would like to be emailed.

I'm using a TypedChoiceField instead of a BooleanField, because BooleanField has weird `required` behaviour. The docs encourage you to set `required=False` if you need to be able to submit a False value, but this means you can't validate that a value was set if you use a RadioWidget.


feedback/models.py
```diff
@@ -34,6 +36,10 @@ class IssueChoices(models.TextChoices):
     created = models.DateTimeField(auto_now_add=True)
     modified = models.DateTimeField(auto_now=True)
 
+    created_by = models.ForeignKey(
+        CustomUser, on_delete=models.CASCADE, blank=False, null=True
```
mitchdawson1982, line 40: Might be worth using the `get_user_model()` helper to refer to the User model. https://docs.djangoproject.com/en/5.1/topics/auth/customizing/#django.contrib.auth.get_user_model
``
from django.contrib.auth import get_user_model

created_by = models.ForeignKey(
        get_user_model(), on_delete=models.CASCADE, blank=False, null=True
)
``

tests/feedback/conftest.py
```diff
@@ -0,0 +1,8 @@
+import pytest
+
+from users.models import CustomUser
```
mitchdawson1982, line 3: Might be worth using the `get_user_model()` helper to save referring to the User model directly. https://docs.djangoproject.com/en/5.1/topics/auth/customizing/#django.contrib.auth.get_user_model
```
from django.contrib.auth import get_user_model

user = get_user_model()
user.objects.create(email="reporter@justice.gov.uk")
```

MatMoore: that's a lot nicer, thanks. I'll refactor this once I've got the thing working

feedback/forms.py
```diff
@@ -53,3 +50,13 @@ class Meta:
...
         formfield_callback = formfield
+
+    send_email_to_reporter = forms.TypedChoiceField(
+        widget=RadioSelect(
+            attrs={"class": "govuk-radios__input", "required": "True"},
+        ),
+        choices=(("Yes", _("Yes")), ("No", _("No"))),
+        coerce=lambda x: x == "Yes",
+        required=True,
+        label=_("Do you want us to email you a copy of this report?"),
```
MatMoore, line 61: not sure about this copy - let me know if you have something better

mitchdawson1982: Perhaps - _Would you like us to email you a copy of this report?_

mitchdawson1982 (dismissed):
LGTM

mitchdawson1982 (approved):
LGTM

PR #1020, opened 2024-11-07
CSV exports

MatMoore:
Add buttons to export metadata as CSV, where we present it in tables.

- [x] Table schema
- [x] List of tables in a database
- [x] List of charts in a dashboard

Screenshot 2024-11-08 at 09 30 55

### Questions for design review
- Is it worth moving the button closer to the data table?
- Does the copy make sense?


LavMatt (commented):
Nice work. Minor comment re. the content of the table csv not including `nullable`

LavMatt (approved):

PR #1030, opened 2024-11-08
feat: add trivy image scanning

MatMoore:
The cloud platform uses Trivy to scan container images for vulnerabilities, while they are running in the cluster. However we should also be scanning the images before we deploy to the cluster.

The kinds of vulnerabilities we are catching here are known vulnerabilities in the packages managed by the operating system in the container. They mostly come from our base images.

This PR runs a trivy scan on each PR.

We send the discovered vulnerabilities to the [code scanning page in github](https://github.com/ministryofjustice/find-moj-data/security/code-scanning?query=is%3Aopen+branch%3Afmj-810-container-scanning), and the "code scanning results" check reports [whether any new vulnerabilities were introduced as part of the PR](https://github.com/ministryofjustice/find-moj-data/pull/1030/checks?check_run_id=32816683933).

It's possible to configure the scan image job to fail depending on whether there are any vulnerabilities at all above a severity threshold (e.g. high or critical), but I don't think this is worth doing right now. I'm expecting we won't be able to fix everything right away, and it would force us to add a lot of exceptions to the ignore file in order to unblock builds.

https://user-guide.cloud-platform.service.justice.gov.uk/documentation/other-topics/trivy-image-scanning.html#trivy-vulnerability-scanning
https://dsdmoj.atlassian.net/wiki/spaces/APPSEC/pages/4988895576/CI+CD+Pipeline+Tooling#Vulnerability-Scan

### DIFF
--- a/.github/workflows/scan-image.yml
+++ b/.github/workflows/scan-image.yml
@@ -1,0 +1,48 @@
+---
+name: Scan Image
+
+on:
+  pull_request:
+    branches:
+      - main
+
+permissions: {}
+
+jobs:
+  scan-image:
+    name: Scan Image
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      security-events: write
+    steps:
+      - name: Checkout
+        id: checkout
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+
+      - name: Build Image
+        id: build_image
+        uses: docker/build-push-action@4f58ea79222b3b9dc2c8bbdd6debcef730109a75 # v6.9.0
+        with:
+          push: false
+          load: true
+          tags: ingestion-notify
+
+      - name: Scan Image (Produce SARIF)
+        id: scan_image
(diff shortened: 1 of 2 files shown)


MatMoore:
> https://user-guide.cloud-platform.service.justice.gov.uk/documentation/other-topics/trivy-image-scanning.html#trivy-vulnerability-scanning

I don't think the PR will be annotated in our case because the scan happens in the deployment workflow. But the issues will appear in the security tab.

I need to check how other teams have done this kind of thing as it seems like it would be better to build the image and scan it as part of the pull request, but I don't see an obvious way to do this.

MatMoore:
Going to rework this following the example in
https://github.com/ministryofjustice/analytical-platform-ingestion-notify/blob/main/.github/workflows/scan-image.yml#L33

We will need a separate build of the image, in order to run the scan pre-merge, but we don't need to push it anywhere.

MatMoore:
Further examples here https://dsdmoj.atlassian.net/wiki/spaces/APPSEC/pages/4988895576/CI+CD+Pipeline+Tooling

murdo-moj (approved):
Looks good - this doesn't change our process day to day but might spawn a few remediation tickets

PR #1040, opened 2024-11-12
Add metadata items to tables and databases (CaDeT)

murdo-moj:
### Added metadata items
* **Audience**: Internal
* ~~**Provider**: Analytical Platform~~ We don't know what provider means. The audience item tells users if they can share data or not.
* **Metadata last updated**
* ~~**Data last updated**~~ Temporarily hidden until the run_results are reliable
* **Refresh schedule**: This is based on available tags in CaDeT
  
### Changed
* I make use now of the timestamp formatter in Django so I have changed timestamps to just be raw as datahub collects them

### Questions for reviewers:
* ~~Is it worth making Audience and Provider an Enum type?~~ I have done this


lib/datahub-client/data_platform_catalogue/entities.py
```diff
@@ -355,6 +348,14 @@ class CustomEntityProperties(BaseModel):
         description="Routes to further information about the data",
         default_factory=FurtherInformation,
     )
+    audience: str = Field(
```
MatMoore, line 351: Can we make this an enum? If it's free text then we'll end up with all sorts and it will be difficult to add conditional logic to the frontend

templates/details_base.html
```diff
@@ -75,6 +75,30 @@ <h2 class="govuk-heading-s govuk-!-margin-top-3">
...
+                {{entity.custom_properties.provider}}
+              </li>
+            {% endif %}
+            {% if entity.last_updated %}
+              <li>
+                <span class="govuk-!-font-weight-bold">Data last updated:</span>
+                {{entity.last_updated|date:"d M Y"}}
+              </li>
+            {% endif %}
+            {% if entity.last_modified %}
```
MatMoore, line 96: #### UX suggestion

"Metadata last updated" feels really out of place here to me - in my mind it's metadata *about metadata* (or data about data about data) whereas everything else in this list is about *the data itself*.

I'm wondering if we could show this in a different way that visually suggests what it applies to, e.g.

mockup with metadata last updated in the bottom right of the panel

We know the concept of metadata is going to confuse a lot of users, so I don't think we should rely on users being able to tell the difference based on labels alone.

lib/datahub-client/data_platform_catalogue/client/graphql_helpers.py
```diff
@@ -215,21 +222,26 @@ def parse_properties(
...
     data_summary = DataSummary.model_validate(custom_properties_dict)
+    tags = parse_tags(entity)
+    expected_refresh_periods = ["daily", "weekly", "monthly"]
+    refresh_period_tags = [
+        tag_ref.display_name
+        for tag_ref in tags
+        if tag_ref.display_name in expected_refresh_periods
+    ]
+    data_summary.refresh_period = " ".join(refresh_period_tags).capitalize()
+    audience = custom_properties_dict.get("Audience", "")
```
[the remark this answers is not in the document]
murdo-moj: We have that convention for tags coming from CaDeT to differentiate them from the other tags used in cadet - however these fields are hardcoded as part of [the ingestion](https://github.com/ministryofjustice/data-catalogue/blob/88edcbcb1a5ac6e4d08a3d0ac2cf6bf4dab9de3c/ingestion/cadet.yaml#L69) and aren't coming from tags in CaDeT. So I'd say no for this.

templates/details_base.html
murdo-moj: I like this idea, I'll implement it 🚀

lib/datahub-client/data_platform_catalogue/client/graphql_helpers.py
```diff
@@ -166,6 +177,28 @@ def parse_tags(entity: dict[str, Any]) -> list[TagRef]:
     return tags
 
 
+def get_refresh_period_from_cadet_tags(
+    tags: list[TagRef],
+    refresh_schedules: list[str] = ["daily", "weekly", "monthly"]
+) -> str:
```
MatMoore, line 183: For Justice Data I'm returning title case for this e.g. "Monthly" etc

Do you want to change your one or shall I change mine?

lib/datahub-client/tests/client/datahub/test_datahub_client.py
```diff
@@ -389,15 +392,15 @@ def test_get_dataset(
...
             },
             domain=DomainRef(display_name="", urn=""),
             governance=Governance(
                 data_owner=OwnerRef(display_name="", email="", urn=""),
                 data_stewards=[],
             ),
             tags=[TagRef(display_name="some-tag", urn="urn:li:tag:Entity")],
-            last_modified=datetime(2024, 3, 5, 6, 16, 47, 814000, tzinfo=timezone.utc),
+            last_modified=1709619407814,
+            provider="LAA",
```
MatMoore, line 403: we've still got provider here, is this being used? (I was assuming we were leaving this out for now)

lib/datahub-client/data_platform_catalogue/entities.py
```diff
@@ -520,6 +525,11 @@ class Table(Entity):
             ]
         ],
     )
+    last_updated: Optional[datetime] = Field(
```
MatMoore, line 528: I've got this the other way round - I've used the existing last_modified for the data last updated. 😕 

What do you think about renaming these fields to `data_last_updated` and `metadata_last_updated`?

templates/details_base.html
```diff
@@ -75,6 +76,18 @@ <h2 class="govuk-heading-s govuk-!-margin-top-3">
                 {{entity.custom_properties.data_summary.refresh_period}}
               </li>
             {% endif %}
+            {% if entity.custom_properties.audience %}
+              <li>
+                <span class="govuk-!-font-weight-bold">Audience:</span>
+                {{entity.custom_properties.audience}}
+              </li>
+            {% endif %}
+            <!-- {% if entity.last_updated %}
```
MatMoore, line 85: This block can be deleted now i think

templates/details_base.html
```diff
@@ -92,6 +105,9 @@ <h2 class="govuk-heading-s govuk-!-margin-top-3">
           </ul>
           {% include "partial/esda_info.html" with is_esda=is_esda %}
         {% endblock metadata_list %}
+        <div class="govuk-body-s" style="text-align: right;">
```
MatMoore, line 108: This will work, but I think it might be worth introducing a CSS class here for reusability.

We already have a custom class called `.summary-card`, so I think we should move that to `scss/components/_summary_card.scss` and add an additional class for this right align:

```scss
.app-summary-card {
    background-color: govuk-colour("light-grey");
    padding: govuk-spacing(3) govuk-spacing(4);
    margin-bottom: govuk-spacing(6);
    overflow-wrap: break-word;
}

.app-summary-card .app-summary-card__footer  {
  text-align: right;
}
```

the above would bring it in line with [BEM naming convention](https://css-tricks.com/bem-101/) we are using elsewhere and the `app-` prefix for custom components.

lib/datahub-client/data_platform_catalogue/client/graphql_helpers.py
```diff
@@ -166,6 +177,28 @@ def parse_tags(entity: dict[str, Any]) -> list[TagRef]:
...
+        schedule
+        for tag_ref in tags
+        for schedule in refresh_schedules
+        if schedule in tag_ref.display_name
+    ]
+    if len(refresh_period_tags) > 1:
+        logger.warn(f"More than one refresh period tag found: {tags=}")
+
+    if refresh_period_tags:
+        refresh_schedule = refresh_period_tags[0]
```
murdo-moj, line 195:
```suggestion
        refresh_schedule = refresh_period_tags[0].capitalize()
```

MatMoore (approved):

PR #1045, opened 2024-11-13
Fmd 984 replace owner refs from app

hjribeiro-moj:
- replaced `data_owner` references from app code
- kept existing references on datahub lib 

keeping PR short to minimize reference problems

it works now in dev:


murdo-moj (approved):
LGTM

PR #1074, opened 2024-11-18
Account for non cadet refresh period

murdo-moj:


lib/datahub-client/data_platform_catalogue/client/graphql_helpers.py
```diff
@@ -249,7 +249,9 @@ def parse_properties(
     usage_restrictions = UsageRestrictions.model_validate(custom_properties_dict)
     data_summary = DataSummary.model_validate(custom_properties_dict)
     tags = parse_tags(entity)
-    data_summary.refresh_period = get_refresh_period_from_cadet_tags(tags)
+    cadet_refresh_period = get_refresh_period_from_cadet_tags(tags)
```
LavMatt, line 252: More a comment on the function `get_refresh_period_from_cadet_tags()` but if multiple refresh tags are found it just uses the first but should it have a preference for returning the most frequent?

LavMatt (dismissed):

LavMatt (approved):

PR #1093, opened 2024-11-25
fmd-1082 - Fix Data Custodian email not being sent to feedback service

hjribeiro-moj:
It is now storing email addr in DB:

image


murdo-moj (approved):
LGTM

PR #1102, opened 2024-11-26
fmd-1081 - Refactor a large il-else block with a mapping

hjribeiro-moj:


lib/datahub-client/data_platform_catalogue/client/search.py
```diff
@@ -107,35 +113,28 @@ def search(
     def _parse_search_results(self, response) -> Tuple[list, list]:
         page_results = []
         malformed_result_urns = []
+
+        EntityMapping = namedtuple('EntityMapping', ['result_type', 'parsing_function'])
```
MatMoore, line 117: I think you can move `EntityMapping` to module level to avoid redefining it each time

lib/datahub-client/data_platform_catalogue/client/search.py
```diff
@@ -107,35 +113,28 @@ def search(
...
-                    )
-                    page_results.append(parsed_result)
-                elif entity_type == "DASHBOARD":
-                    parsed_result = self._parse_container(
-                        entity, matched_fields, ResultType.DASHBOARD
-                    )
-                    page_results.append(parsed_result)
-                else:
-                    raise Exception
+                parsed_result = entity_types_map[entity_type].parsing_function(entity, matched_fields, entity_types_map[entity_type].result_type)
```
MatMoore, line 133: Since there's a lot going on in this search class, I'm wondering if it's worth going a step further and extracting out helper classes for parsing the various entities.

something like

```python
class Parser:
  def parse_types_and_subtypes(self): ...

class ContainerParser(Parser):
  def parse(self, entity, matched_fields, result_type): ...
  
class DatasetParser(Parser):
  def parse(self, entity, matched_fields, result_type): ...
```

In that case

- `_parse_dataset`, `_parse_container` and `_parse_types_and_subtypes` become public methods of the new classes, and can be unit tested
- instead of a parsing function, you map to an instance of one of these classes, and call its `parse` method

what do you think?

MatMoore (approved):

lib/datahub-client/data_platform_catalogue/client/search.py
murdo-moj: This is a nice idea! We've moved all these changes to the larger refactor PR where the mapping is a bit more complicated, perhaps we could work this in

murdo-moj:
We address these changes in #1112

PR #1110, opened 2024-11-27
refactor: rename timestamps

MatMoore:
There are various datetimes returned by Datahub, and it's easy to confuse them. It seems like we are not always using the right one, so as a first step I'm renaming all the fields and parse functions to clearly describe what they are actually returning.

I've also exposed Datahub's lastModified value as I think this is the one we should be using for data last updated.

> A timestamp documenting when the asset was last modified in the source Data Platform (not on DataHub)

This does not change what is displayed yet, I'll do that in a follow up PR.

### DIFF
--- a/lib/datahub-client/data_platform_catalogue/client/datahub_client.py
+++ b/lib/datahub-client/data_platform_catalogue/client/datahub_client.py
@@ -40,19 +40,19 @@ from data_platform_catalogue.client.exceptions import (
 )
 from data_platform_catalogue.client.graphql_helpers import (
     parse_columns,
-    parse_created_and_modified,
     parse_custodians,
+    parse_data_created,
     parse_data_owner,
     parse_domain,
     parse_glossary_terms,
-    parse_last_modified,
+    parse_last_datajob_run_date,
+    parse_metadata_last_ingested,
     parse_names,
     parse_properties,
     parse_relations,
     parse_stewards,
     parse_subtypes,
     parse_tags,
-    parse_updated,
 )
 from data_platform_catalogue.client.search import SearchClient
 from data_platform_catalogue.entities import (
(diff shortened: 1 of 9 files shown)


lib/datahub-client/data_platform_catalogue/client/datahub_client.py
```diff
@@ -275,9 +275,9 @@ def get_table_details(self, urn) -> Table:
             custodians = parse_custodians(response)
             tags = parse_tags(response)
             glossary_terms = parse_glossary_terms(response)
-            created, modified = parse_created_and_modified(properties)
-            modified = parse_last_modified(response)
-            updated = parse_updated(response)
+            created = parse_data_created(properties)
+            modified = parse_metadata_last_ingested(response)
+            updated = parse_last_datajob_run_date(response)
```
LavMatt, line 280: I think this works at the moment - but could there be tables that we want to get the lastModified date from the property in datahub, like when a table is not produced by dbt?

LavMatt (commented):
Do you think we could go a bit further and rename the properties in the pydantic entity model. Every entity should(or could) have a `data_last_modified` and a `metadata_last_updated` property so we could put these in the base entity class and remove reference to them elsewhere in entities.py? Plus they'd be clearer when used in the find-moj-data code

lib/datahub-client/data_platform_catalogue/client/datahub_client.py
MatMoore: This is one of the bits I think is actually wrong - but I'm saving that for the next PR, this one is just renaming things

MatMoore:
> Do you think we could go a bit further and rename the properties in the pydantic entity model. Every entity should(or could) have a `data_last_modified` and a `metadata_last_updated` property so we could put these in the base entity class and remove reference to them elsewhere in entities.py? Plus they'd be clearer when used in the find-moj-data code

Yeah that makes sense... I'll have a look at this now

lib/datahub-client/data_platform_catalogue/client/search.py
```diff
@@ -422,7 +422,7 @@ def _parse_container(
         """
         tags = parse_tags(entity)
         terms = parse_glossary_terms(entity)
-        last_modified = parse_last_modified(entity)
+        last_modified = parse_metadata_last_ingested(entity)
```
MatMoore, line 425: there is inconsistency here too but I'm not sure we're doing anything with this value in the frontend

murdo-moj (approved):

PR #1135, opened 2024-12-04
refactor: use type mapping objects

murdo-moj:
* Use `FindMoJDataEntityMapper` objects to translate from FMD to Datahub types and subtypes
* I chose instances of this class rather than an object to represent our entity types as there's no logic we need to group with the mappings. Also instantiating classes all over the code was messy.
* Add `FindMoJDataEntityType` Enum
* Remove `EntityTypes`
* Remove `SearchClient.fmd_type_to_datahub_types_mapping`
* The formation of `entity_type_filters` is now nicer.

Note subtype "Metric" is not included in any of our searches now

### DIFF
--- a/home/forms/search.py
+++ b/home/forms/search.py
@@ -1,8 +1,8 @@
 from copy import deepcopy
 from urllib.parse import urlencode
 
+from data_platform_catalogue.entities import FindMoJDataEntityType
 from data_platform_catalogue.search_types import DomainOption
-from data_platform_catalogue.entities import EntityTypes
 from django import forms
 from django.utils.translation import gettext as _
 
@@ -38,7 +38,7 @@ def get_entity_types():
     return sorted(
         [
             (entity.name, entity.value)
-            for entity in EntityTypes
+            for entity in FindMoJDataEntityType
             if entity.name != "GLOSSARY_TERM"
         ]
     )
--- a/home/forms/search.py
+++ b/home/forms/search.py
@@ -1,7 +1,7 @@
 from copy import deepcopy
 from urllib.parse import urlencode
 
-from data_platform_catalogue.entities import FindMoJDataEntityType
+from data_platform_catalogue.entities import FindMoJdataEntityType
 from data_platform_catalogue.search_types import DomainOption
 from django import forms
 from django.utils.translation import gettext as _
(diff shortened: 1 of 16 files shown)


home/service/details.py
```diff
@@ -258,12 +261,12 @@ def _get_context(self):
...
         return {
             "entity": self.publication_dataset_metadata,
-            "entity_type": EntityTypes.PUBLICATION_DATASET.value,
+            "entity_type": PublicationDatasetEntityMapper.find_moj_data_type,
             "parent_entity": self.parent_entity,
             "platform_name": friendly_platform_name(
                 self.publication_dataset_metadata.platform.display_name
             ),
-            "parent_type": EntityTypes.PUBLICATION_COLLECTION.name.lower(),
+            "parent_type": DatabaseEntityMapper.datahub_type.lower(),
```
MatMoore: I think we can drop the `lower()` here, and change templates accordingly. We're not surfacing datahub types to users, so the case shouldn't matter as long as we are consistent.

That said, if we do a bigger refactor, then I think this is an example where we can eliminate the Datahub type entirely and replace it with our own entity type. The end state of that work should be that the mapping from Datahub to Find MoJ data entity types is done entirely within the `lib/datahub-client` modules, and Find MoJ data does not reference datahub entity types at all (as @hjribeiro-moj suggested last week)

home/forms/search.py
```diff
@@ -1,8 +1,8 @@
 from copy import deepcopy
 from urllib.parse import urlencode
 
+from data_platform_catalogue.entities import FindMoJDataEntityType
```
hjribeiro-moj, line 4: I am not sure we need to prefix with `FindMoJData`. that is the app name, the lib should be reusable independent of what the app is called.

if an app client has a specific data model, then it should implement its own entities map and convert from `EntityTypes` to it.

lib/datahub-client/data_platform_catalogue/entities.py
```diff
@@ -42,92 +52,54 @@ class FindMoJDataEntityMapper:
...
-        "publication_collection",
-    )
-
-    def __new__(cls, value, datahub_entity_type, url_formatted):
-        obj = object.__new__(cls)
-        obj._value_ = value
-        obj.datahub_entity_type = datahub_entity_type
-        obj.url_formatted = url_formatted
-        return obj
+TableEntityMapper = FindMoJDataEntityMapper(
```
MatMoore, line 55: I wonder if `Mapping` (i.e. a noun) would be a better name than `Mapper` here. Since it's holding data rather than doing the mapping.

MatMoore (approved):
Nice, this looks like an improvement from what we had before. Just a couple of minor suggestions.

hjribeiro-moj (approved):
LGTM

PR #1148, opened 2024-12-06
Add liveness & readiness probe for zero downtime deployments

mitchdawson1982:
I initially went with an approach of a dedicated `/health` endpoint for the probes to target, however probes consider any http status code between 200 - 399 as valid. When a http request targets the home page `/`, the application should attempt authentication redirection with a status of `302`, hence this is valid and exposes a greater level of application functionality.

- defines liveness probe to check for application lock ups and restarts
- defines readiness probe to check for application availability for zero downtime deployments


MatMoore:
Nice, thanks for sorting this 👍🏻

Issue #1108, opened 2024-11-27
"You are now leaving Find MoJ data" transitionary page


murdo-moj:
### User story

As a user I want to be notified when I am leaving Find MoJ data to another gov.uk design pattern styled website.

### Context

Other govuk services we link to eg data.gov.uk have very similar styling. This makes it hard for the user to notice that they have left the service.

### Ideas

Make a transitionary page which informs users they are leaving FMD

YvanMOJdigital:
check for moj design system component

LavMatt:
This design as is with an auto redirect is bad for accessibility. I have reached out to the design slack channel in gov digital slack
https://ukgovernmentdigital.slack.com/archives/C062GAGLW/p1733825790263679

If we want to keep it we likely need to add a second click rather than automatically redirect

joe-horton-moj:
Comments and recommendation left in [PR conversation](https://github.com/ministryofjustice/find-moj-data/pull/1153#pullrequestreview-2511367479).

**TL:DR** – don't implement a transitionary page for external links and monitor the issue in future UR sessions.

LavMatt:
> Comments and recommendation left in [PR conversation](https://github.com/ministryofjustice/find-moj-data/pull/1153#pullrequestreview-2511367479).
> 
> **TL:DR** – don't implement a transitionary page for external links and monitor the issue in future UR sessions.
> 
> What are your thoughts [@LavMatt](https://github.com/LavMatt)? I think you came to same reasoning last time we discussed in on stand-up?

Yeah totally agree with your comments Joe. If it continues to be a problem arise in future UR, we should explore solutions other than creating a transitionary page around design and stronger brand identity within our service.

I shall close the PR

PR #1153, opened 2024-12-09
feat: add transitionary notification redirect page for external urls

LavMatt:
PR for #1108 

This is in response to some UR where users felt it was not made clear enough that certain links were taking them outside of the find moj data service.

## What i've done
- created a redirect page template
- created a redirect view
- changed external links in other templates to call the new view and pass the actual external link as a query parameter in the request, which then renders the redirect template which redirects

## Accessibility
My initial go at this didn't play to well with assistive tech in the form of the mac's voiceover screen reader. To work around this i have made the new redirect page autofocus on the heading with the main info so that it's read out first and added additional info around where the user is being directed to, to the aria-label.

And on the subject of accessibility i spotted a mistake that had slipped through where our links on some pages didn't include `(opens in new tab)` in the link text. I have fixed that issue as part of this PR too.


joe-horton-moj (commented):
After reviewing the slack thread, my initial thought is that this solution could make the user experience more tedious and ultimately won't solve the core issue. 

As I understand correctly, we had a few users notice that they weren't on "Find MoJ data" anymore after clicking on certain external links and were left confused wondering where they were, not noticing they were on a different GOV service.

This feels like a symptom of a wider issue (touched on in the thread) about the similarity in look and feel of GOV services and the difficulty to distinguish the service from initial glance.

My opinion at this stage would be park the bus on implementing anything and continue to monitor the issue in future UR sessions.

If we find the issue coming up again, I'd recommend two different options:

1. design and test a different styling / interaction of our external links
2. developing a stronger brand identity or look and feel for "Find MoJ data"

Issue #1097, opened 2024-11-25
Make sure all components are documented


MatMoore:
The purpose of this ticket is to make sure that we "Maintain up-to-date documentation for all implemented components, APIs, and dependencies to facilitate smooth collaboration and knowledge transfer within MOJ Digital"

We should make sure every component has a readme or other documentation, and remove anything that is out of date. Consider moving content into the runbooks where it makes sense.

- [x] [Find MOJ data repo](https://github.com/ministryofjustice/find-moj-data/blob/main/README.md)
- [x] [Custom frontend components](https://github.com/ministryofjustice/find-moj-data/blob/main/scss/components/README.md)
- [x] [The custom ingestion sources / transformers](https://github.com/ministryofjustice/data-catalogue/tree/main/ingestion)
- [ ] [Data catalogue repo](https://github.com/ministryofjustice/data-catalogue/blob/main/README.md)

LavMatt:
mostly updated and reviewed.

Outstanding is the data catalogue repo [readme secrets section](https://github.com/ministryofjustice/data-catalogue/tree/main?tab=readme-ov-file#managing-database-secrets) - Currently has a circular flow for some postgres and opensearch secrets which may be unnecessary now. Will need further investigating and potentially streamlining and then the readme here should be updated accordingly

PR #1171, opened 2024-12-13
fix: properly set is nullable column waffle switch

LavMatt:
part of [#1097](https://github.com/ministryofjustice/find-moj-data/issues/1097)

The switch was missing from the app-entrypoint.sh script and was not documented in the readme


murdo-moj (approved):
Good spot 👍

PR #1173, opened 2024-12-13
Urlencode entity url for notifications

mitchdawson1982:
- Adds `encoded_entity_url` method to the `Issue` model that returns a url encoded entity url, and associated test. This resolves the issue with the notify templates incorrect rendering urls with `)` within them.


feedback/models.py
```diff
@@ -55,3 +57,19 @@ class IssueChoices(models.TextChoices):
     entity_name = models.CharField(max_length=250)
     entity_url = models.CharField(max_length=250)
     data_custodian_email = models.CharField(max_length=250)
+
+    @property
+    def encoded_entity_url(self):
+        parsed_url: ParseResult = urlparse(self.entity_url)
+        encoded_path: str = quote(parsed_url.path)
```
MatMoore, line 64: This seems ok as long as you've tested that the generated URLs still work ok 👍🏻 

It seems a bit odd to have to reencode here, as I was expecting this to be handled by django's `url` tag and `reverse` functions, which [encode the result](https://docs.djangoproject.com/en/5.1/ref/urlresolvers/#reverse:~:text=The%20string%20returned%20by%20reverse()%20is%20already%20urlquoted.%20For%20example%3A), but it seems like there is a difference in behaviour here

```python
>>> from urllib.parse import quote
>>> quote('foo(bar)')
'foo%28bar%29'
```

```python
>>> from django.utils.encoding import iri_to_uri
>>> iri_to_uri('foo(bar)')
'foo(bar)'
```

Apparently encoding parentheses is optional in a URI, so either should be valid normally.

MatMoore (approved):

feedback/models.py
mitchdawson1982: As discussed it is done here to avoid changing existing url behaviour. The only case we are catering for is when the url is sent to the notify service.

PR #1196, opened 2024-12-18
Remove i18n

MatMoore:
We decided to disable i18n again as we're not using it for anything other than reusing messages in multiple templates, and it was more trouble than it was worth.


mitchdawson1982 (approved):
LGTM

PR #1210, opened 2025-01-07
Enhance GraphQL queries to includeSoftDelete parameter for relationships

murdo-moj:
* We're getting [errors](https://mojdt.slack.com/archives/C07SFN0MHRP/p1736258081094319) to do with the caseman database, as table descriptions are empty. There's a related issue which is that soft deleted tables are being listed in the database's relationships, which are removed by stateful ingestion
* This fix sets an empty string as a description for all tables in a database where there is no description
* The parameter to exclude soft deleted relationships is also there, but it doesn't seem to work

### DIFF
--- a/lib/datahub-client/data_platform_catalogue/client/parsers.py
+++ b/lib/datahub-client/data_platform_catalogue/client/parsers.py
@@ -430,9 +430,12 @@ class EntityParser:
                     if relation.get("entity", {}).get("properties") is not None
                     else relation.get("entity").get("name", "")
                 )
+                properties = relation.get("entity", {}).get("properties")
+                description_field = relation.get("entity").get("properties", {}).get("description", "")
                 description = (
-                    relation.get("entity").get("properties", {}).get("description", "")
-                    if relation.get("entity", {}).get("properties") is not None
+                    description_field
+                    if properties is not None
+                    and description_field is not None
                     else ""
                 )
                 tags = self.parse_tags(relation.get("entity"))
--- a/lib/datahub-client/data_platform_catalogue/client/parsers.py
+++ b/lib/datahub-client/data_platform_catalogue/client/parsers.py
@@ -47,8 +47,6 @@ DATA_CUSTODIAN = "urn:li:ownershipType:data_custodian"
 
 
 class EntityParser:
-    def __init__(self):
-        pass
 
     def parse(self, search_response) -> SearchResult:
         """Parse graphql response to a SearchResult object"""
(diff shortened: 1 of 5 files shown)


lib/datahub-client/data_platform_catalogue/client/parsers.py
```diff
@@ -430,9 +430,12 @@ def parse_relations(
                     if relation.get("entity", {}).get("properties") is not None
                     else relation.get("entity").get("name", "")
                 )
+                properties = relation.get("entity", {}).get("properties")
+                description_field = relation.get("entity").get("properties", {}).get("description", "")
```
MatMoore, line 434: It looks like this code fails on some cases that the previous one doesn't, for example `{"entity": {"properties": None}}` triggers `AttributeError: 'NoneType' object has no attribute 'get'`

Elsewhere in this module we have expressions like `entity.get("properties") or {}` to handle both missing and None values. There might be a cleaner way than this but it should remove the edge cases.

If it helps, I wrote a hypothesis test (with some help from ChatGPT):

```python
from hypothesis import given
from hypothesis import strategies as st

from parser_change import after, before

strategy = st.fixed_dictionaries(
    {
        "entity": st.one_of(
            st.just({}),
            st.fixed_dictionaries(
                {
                    "properties": st.one_of(
                        st.none(),
                        st.dictionaries(
                            keys=st.sampled_from(["description", "banana"]),
                            values=st.one_of(st.none(), st.text()),
                        ),
                    )
                }
            ),
        )
    }
)


@given(strategy)
def test_equivalent(s):
    try:
        previous_value = before(s)
    except Exception:
        pass
    else:
        if previous_value is None:
            assert after(s) == ""
        else:
            # New version should be equivalent to old version
            assert previous_value == after(s)
```

murdo-moj: I've changed the implementation to account for properties being filled as none, and description being filled as none. In both cases we want a value of "" for description

lib/datahub-client/data_platform_catalogue/client/parsers.py
```diff
@@ -430,11 +428,11 @@ def parse_relations(
...
                     else relation.get("entity").get("name", "")
                 )
-                description = (
-                    relation.get("entity").get("properties", {}).get("description", "")
-                    if relation.get("entity", {}).get("properties") is not None
-                    else ""
-                )
+                properties = relation.get("entity", {}).get("properties")
+                if properties is not None:
+                    description = properties.get("description") or ""
```
mitchdawson1982, line 433:
```suggestion
                    description = properties.get("description", "")
```

lib/datahub-client/data_platform_catalogue/client/parsers.py
```diff
@@ -430,11 +428,11 @@ def parse_relations(
...
+                elif properties is None:
+                    description = ""
```
mitchdawson1982, line 435: We have tested for not None in the `if`
```suggestion
                else:
                    description = ""
```

lib/datahub-client/data_platform_catalogue/client/parsers.py
murdo-moj: sometimes we have {"description": None}, which we want to resolve to `""`, and would with this change resolve to `None`

mitchdawson1982 (approved):

PR #1229, opened 2025-01-13
refactor: remove upsert_* methods and stop treating data_platform_catalogue as a separate package

MatMoore:
We no longer depend on this library for ingestion, so we can remove all the unused methods, and the separate pyproject.toml.

I've also taken the opportunity to rename and flatten the package structure a bit, so we have `datahub_client` rather than `data_platform_catalogue.client` (the Data platform naming predates Find MoJ data)


hjribeiro-moj (approved):
LGTM

PR #1231, opened 2025-01-14
Add timestamp validation and associated test

mitchdawson1982:
- Utilises the Annotated pattern for field validation which delegates to a callable. https://docs.pydantic.dev/latest/concepts/validators/#field-validators

### DIFF
--- a/tests/datahub_client/conftest.py
+++ b/tests/datahub_client/conftest.py
@@ -1,25 +1,8 @@
-from datetime import datetime, timedelta
 from pathlib import Path
 from typing import Any, Dict
 
 import pytest
 from datahub.metadata.schema_classes import DomainPropertiesClass
-
-from datahub_client.entities import (
-    AccessInformation,
-    CustomEntityProperties,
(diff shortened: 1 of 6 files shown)


tests/datahub_client/conftest.py
```diff
@@ -50,3 +67,85 @@ def pytest_addoption(parser):
         default=False,
     )
     parser.addoption("--copy-output-files", action="store_true", default=False)
+
+
+@pytest.fixture
+def entity_data_with_timestamps_in_future():
```
MatMoore, line 73: I think it might be worth inlining this fixture into the test, as it's probably not going to be used anywhere else?

mitchdawson1982: I originally had it inline but moved it to keep the test cleaner given the size of the data structure. Happy to move it back.

hjribeiro-moj (approved):
LGTM

hjribeiro-moj (approved):
LGTM

PR #1233, opened 2025-01-14
Use tags to populate subject areas

MatMoore:
Now that the ingestion code is setting tags in DataHub for the subject areas, we can start to use these for search/browse instead of the DataHub domain.

This work will allow us to assign entities to more than one subject area, so we don't need to worry about something not precisely fitting into a single category when redefining the taxonomy.

## Known issues
I've used the test environment for testing this since we've had some OpenSearch issues on the dev environment. When I compared before and after I noticed a few discrepancies, but the number of entities in each subject area is almost the same.

It seems like the discrepancies are due to two things, neither of which seem like blockers:

1. Entries in the catalogue that are no longer being updated, but haven't been deleted automatically as part of a stateful ingestion, no longer show up in a subject area ([example](https://datahub-catalogue-test.apps.live.cloud-platform.service.justice.gov.uk/dataset/urn:li:dataset:(urn:li:dataPlatform:dbt,cadet.awsdatacatalog.use_of_force.cte_post_may_stg,PROD)/Schema?is_lineage_mode=false&schemaFilter=)). These entries should be deleted altogether but for whatever reason they haven't been. This affects CaDeT databases, and some ESDAs (to be expected as these do not belong to an ingestion).
2. When ingesting CaDeT databases, there are cases where the database contains models that are in different domains. We used to pick a single domain to use for the database, but now we assign tags for all of them ([example](https://datahub-catalogue-test.apps.live.cloud-platform.service.justice.gov.uk/container/urn:li:container:23005579d7a83704d768231acea74362/Entities?is_lineage_mode=false)). Such databases now show up in multiple subject areas. I hadn't intended to change this right away but I think it is the behaviour we ultimately want.

## Next steps

- Currently where there are multiple subjects tags assigned to an entity we just pick the first one. I'm planning to change this to show all of them. At the same time I can look into the css bug that's making it look wonky when the subject area tag flows onto a new line.
- There is likely some further refactoring that can be done in the `home` app in Find MoJ data as well, e.g. some remaining references to "domains" that can be cleaned up.


datahub_client/parsers.py
```diff
@@ -96,26 +97,29 @@ def parse_tags(entity: dict[str, Any]) -> list[TagRef]:
...
-            display_name = domain_properties.get("name", "")
-        else:
-            display_name = ""
+    def parse_subject_area(self, entity: dict[str, Any]) -> DomainRef:
+        outer_tags = entity.get("tags") or {}
+        for tag in outer_tags.get("tags", []):
+            tag_inner = tag["tag"]
+            properties = tag.get("properties") or {}
+            name = properties.get("name") or tag_inner.get("name")
+            subject_area = SubjectAreaTaxonomy.get_top_level(name)
```
murdo-moj, line 116:
```suggestion
            # Only tags included in the top level taxonomy are collected
            subject_area = SubjectAreaTaxonomy.get_top_level(name)
```

datahub_client/parsers.py
```diff
@@ -96,26 +97,29 @@ def parse_tags(entity: dict[str, Any]) -> list[TagRef]:
...
+            if subject_area is None:
+                continue
+            else:
+                return DomainRef(display_name=name, urn=subject_area.urn)
 
-        return DomainRef(display_name=display_name, urn=domain_id)
+        return DomainRef(display_name="", urn="")
```
murdo-moj, line 122: Renaming `DomainRef` to `SubjectAreaRef` comes later?

datahub_client/search/search_client.py
```diff
@@ -126,45 +124,54 @@ def _parse_search_results(self, response) -> Tuple[list, list]:
...
-        """
-        Returns domains that can be used to filter the search results.
-        """
+    ):
+
         formatted_filters = map_filters(filters)
         formatted_filters = formatted_filters[0]["and"]
-        variables = {
+
+        variables = variables = {
```
murdo-moj, line 137:
```suggestion
        variables = {
```

datahub_client/search/search_client.py
```diff
@@ -126,45 +124,54 @@ def _parse_search_results(self, response) -> Tuple[list, list]:
...
-            name = properties.get("name", "")
-            entities = domain.get("entities", {})
-            total = entities.get("total", 0)
+            subject_areas.append(
+                SubjectAreaOption(subject_area.domain_urn, name, count)
+            )
 
-            list_domain_options.append(SubjectAreaOption(urn, name, total))
-        return list_domain_options
+        return sorted(subject_areas, key=lambda s: s.name)
```
murdo-moj, line 174:
```suggestion
        sorted_subject_areas = sorted(subject_areas, key=lambda s: s.name)
        
        return sorted_subject_areas
```

home/service/search.py
```diff
@@ -88,7 +88,9 @@ def _get_search_results(self, page: str, items_per_page: int) -> SearchResponse:
 
         filter_value = []
         if subject_area:
-            filter_value.append(MultiSelectFilter("domains", [subject_area]))
+            filter_value.append(
+                MultiSelectFilter("tags", [subject_area.replace(":domain:", ":tag:")])
```
murdo-moj, line 92: Did you consider changing the form so you don't need to do this replace? Or the graphQL query which populates the form options for subject area?

datahub_client/search/search_client.py
hjribeiro-moj: why do we need the extra variable?

datahub_client/parsers.py
```diff
@@ -96,26 +97,29 @@ def parse_tags(entity: dict[str, Any]) -> list[TagRef]:
...
-        inner_domain = domain.get("domain") or {}
-        domain_id = inner_domain.get("urn", "")
-        if inner_domain:
-            domain_properties, _ = self.parse_properties(inner_domain)
-            display_name = domain_properties.get("name", "")
-        else:
-            display_name = ""
+    def parse_subject_area(self, entity: dict[str, Any]) -> DomainRef:
+        outer_tags = entity.get("tags") or {}
+        for tag in outer_tags.get("tags", []):
```
hjribeiro-moj, line 112: do we need the `[]` as a default?

datahub_client/search/search_client.py
```diff
@@ -126,45 +124,54 @@ def _parse_search_results(self, response) -> Tuple[list, list]:
...
+        """
+        Iterate over all the tag values, and return values for those
+        that match the top level subject areas.
+        """
+        subject_areas: list[SubjectAreaOption] = []
+
+        for aggregation in facets[0]["aggregations"]:
+            count = aggregation["count"]
+            entity = aggregation["entity"]
+            properties = entity.get("properties") or {}
```
hjribeiro-moj, line 164: we don't need `entity` and ` properties` variables, as they are used in the following lines only, but happy to let them on if we feel it improves readability.

To me they hint that they're going to be used further down.

datahub_client/parsers.py
MatMoore: Yeah I'm gonna sort that out next, instead of returning one DomainRef, it should return any number of subject areas

datahub_client/search/search_client.py
MatMoore: we don't! think I left it in from some earlier testing

home/service/search.py
MatMoore: I'm gonna keep this in for now just in case anyone has links bookmarked with the old filters applied. But I will update the URLs in the next step so it uses the tag URNs instead.

datahub_client/parsers.py
MatMoore: we might be able to swap for `outer_tags["tags"]` if it's guaranteed to always be set, but I just wanted to be defensive here because we've run into a lot of issues with this API returning None when a value is expected

datahub_client/search/search_client.py
MatMoore: will take a look at this 👀

MatMoore: I've extracted this out into another parser function

datahub_client/parsers.py
hjribeiro-moj: my bad 🤦

hjribeiro-moj (approved):
LGTM

PR #1247, opened 2025-01-16
fix: remove left margin on tags that flow onto a new line

MatMoore:
Search results with long titles currently render strangely because the tag always has a left margin, but it can wrap onto a new line.

Workaround this by using flexbox to align the title and the tag. 

## Before
Screenshot 2025-01-16 at 16 49 48

## After
Screenshot 2025-01-16 at 17 02 39

### DIFF
--- a/scss/base.scss
+++ b/scss/base.scss
@@ -64,8 +64,15 @@ code {
   min-width: 5.5em;
 }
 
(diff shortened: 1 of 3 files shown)


MatMoore:
Closing this as it's likely to be redundant soon

PR #1257, opened 2025-01-22
feat: embed quicksight meatdata quality dashboard

LavMatt:


mitchdawson1982 (approved):
Nice 👍🏻

dashboard/views.py
```diff
@@ -0,0 +1,20 @@
...
+from django.http import HttpResponse
+from django.shortcuts import render
+
+from .service import generate_embed_url_for_anonymous_user
+
+
+def metadata_quality_dashboard(request) -> HttpResponse:
+    quicksight_embedded_url = generate_embed_url_for_anonymous_user()
+    if not quicksight_embedded_url:
+        raise BadRequest("Error generating QuickSight embedded URL")
```
MatMoore, line 11: This seems like the wrong use of BadRequest, since the request was valid, it's just the server is misconfigured. So I think we should technically be returning a 500 here, not a 400.

MatMoore (dismissed):
LGTM, just noticed a small issue with the exception handling.

tests/core/test_middleware.py
```diff
@@ -23,7 +23,7 @@ def test_middleware_renders_bad_request_response():
     get_response = MagicMock()
     request = MagicMock()
     middleware = CustomErrorMiddleware(get_response)
-    error = BadRequest()
+    error = BadRequest("test")
```
murdo-moj, line 26: is this meant to say test?

MatMoore (approved):

PR #1303, opened 2025-01-28
💬 Update audience property

mitchdawson1982:
### Summary
Changes the audience property name to classification and applies the appropriate security classification and additional marking where appropriate based on [this](https://www.gov.uk/government/publications/government-security-classifications/government-security-classifications-policy-quick-read-html) guidance.

`Published` becomes `Official`
`Internal` becomes `Official-Sensitive` (Security Classification-Additional Marking)

- Defines new `Classification` entity
- Changes audience property to classification
- Updates tests

### DIFF
--- a/datahub_client/entities.py
+++ b/datahub_client/entities.py
@@ -114,9 +114,19 @@ Mappers = [
 ]
 
 
-class Audience(Enum):
-    INTERNAL = "Internal"
-    PUBLISHED = "Published"
+class Classification(Enum):
+    """Enumeration representing the security classification of the data.
+    Attributes:
+        OFFICIAL (str): Represents data classified as "Official".
+        OFFICIAL_SENSITIVE (str): Represents data classified as "Official-Sensitive".
+    Notes:
+        This enumeration replaces the Audience entity.
+        - "Published" becomes "Official".
+        - "Internal" becomes "Official-Sensitive".
+    """
+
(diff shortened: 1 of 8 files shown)


datahub_client/entities.py
```diff
@@ -114,9 +114,19 @@ class FindMoJdataEntityMapper:
...
+    Attributes:
+        OFFICIAL (str): Represents data classified as "Official".
+        OFFICIAL_SENSITIVE (str): Represents data classified as "Official-Sensitive".
+    Notes:
+        This enumeration replaces the Audience entity.
+        - "Published" becomes "Official".
+        - "Internal" becomes "Official-Sensitive".
+    """
+
+    OFFICIAL = "Official"
```
MatMoore, line 128: Do we need to capitalise these for consistency with https://www.gov.uk/government/publications/government-security-classifications/government-security-classifications-policy-html?

I'm also wondering if we should use "OFFICIAL - FOR PUBLIC RELEASE" - which is given as an example of a handling instruction.

MatMoore (approved):

datahub_client/entities.py
mitchdawson1982: Handling instructions were beyond the scope of the ticket this time around, however ill raise a ticket for consideration. We have followed the DBT approach of lower casing the value.

PR #1308, opened 2025-01-29
feat: add new subject areas behind a switch

MatMoore:
Once we're happy things have been tagged to the new subject areas in prod, we can switch this on. This switch is independent of the new homepage design.

This extra code should be removed once this is live, I've just added it to help with the release.


scripts/app-entrypoint.sh
```diff
@@ -14,6 +14,7 @@ python manage.py migrate
 python manage.py waffle_switch search-sort-radio-buttons off --create # create switch with default setting
 python manage.py waffle_switch display-result-tags off --create # create display tags switch with default off
 python manage.py waffle_switch show_is_nullable_in_table_details_column off --create # create isnullable column switch with default off
+python manage.py waffle_switch new_subject_areas off --create
```
hjribeiro-moj, line 17: should we add a comment to delete `ALL_SUBJECT_AREAS_OLD` once we move on definitely from this toggle?

hjribeiro-moj (approved):
LGTM

Issue #1295, opened 2025-01-27
Move Access this data section


teeceeas:
![Image](https://github.com/user-attachments/assets/7dd4a2c1-e4ff-4986-a9de-7233ba3f9c86)

Issue #1305, opened 2025-01-28
Consider using handling instructions alongside security classifications


mitchdawson1982:
### User Story

As a developer I want to make it easy for users to understand the scope of the security classification property.

### Value / Purpose

With the use of security classifications, most of our data is marked `Official`. It is not obvious that this classification means that the data can be made publicly available. 

If we were to utilise a handling instruction such as `FOR PUBLIC RELEASE` alongside the classification, this would make it clear that the data could be distributed publicly for example.

Guidance - https://www.gov.uk/government/publications/government-security-classifications/government-security-classifications-policy-html#additional-markings

### Proposal

We could utilise handling instructions to provide further context to ambiguous security classifications.

Example:

![Image](https://github.com/user-attachments/assets/06577211-1051-47c7-a785-2daead279ddf)

Current - 

> Security classification: Official

Proposed - 

> Security classification: Official-For Public Release

### Additional Information

- [GSC](https://www.gov.uk/government/publications/government-security-classifications/government-security-classifications-policy-quick-read-html)

Handling Instructions | OFFICIAL | SECRET | TOP SECRET | OFFICIAL (with –SENSITIVE marking)
-- | -- | -- | -- | --
RECIPIENTS ONLY | YES (WITH  ‘-SENSITIVE’ MARKING ONLY) | YES | YES | YES
FOR PUBLIC RELEASE | YES | NO | NO | NO
ORGANISATION USE ONLY | YES | YES | YES | YES
HMG USE ONLY | YES | YES | NO | YES
EMBARGOED | YES | YES | NO | YES
Descriptors |   |   |   |  
PERSONAL DATA | YES | YES | YES | YES
LEGAL PROFESSIONAL PRIVILEGE (LPP) | YES | YES | YES | YES
LEGAL | YES | NO | NO | YES
MARKET SENSITIVE | YES | YES | YES | YES
COMMERCIAL | YES | YES | YES | YES
HR / MANAGEMENT | YES | YES | YES | YES

### Definition of Done

TBC

joe-horton-moj:
Thanks for providing that GOV link guidance on HANDLING INSTRUCTIONS.

I didn't know there were additional instructions beyond the security classification.

A couple of them make sense and are understandable. There are quite a few there where as a non-technical person – I wouldn't understand what that classification meant for me and how I access the data?

joe-horton-moj:
[...]

PR #1317, opened 2025-01-30
feat: Hide request access contact section for publications

murdo-moj:
* Hide Request Access section for non-sensitive data, where there is already an "Access this data" section
* The other point on the ticket is to change the "Access this data" to an h2 - this is already done

https://github.com/ministryofjustice/find-moj-data/issues/1295

image

### DIFF
--- a/templates/partial/contact_info.html
+++ b/templates/partial/contact_info.html
@@ -1,19 +1,21 @@
 <div class="govuk-body">
-  <h2 class="govuk-heading-s govuk-!-margin-bottom-1">Request access</h2>
-  <p id="request_access" class="govuk-body">
+  {% if entity_type != "Publication collection" and entity_type != "Publication dataset" %}
+    <h2 class="govuk-heading-s govuk-!-margin-bottom-1">Request access</h2>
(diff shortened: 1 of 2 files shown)


templates/partial/contact_info.html
```diff
@@ -1,19 +1,21 @@
 <div class="govuk-body">
-  <h2 class="govuk-heading-s govuk-!-margin-bottom-1">Request access</h2>
-  <p id="request_access" class="govuk-body">
+  {% if entity_type != "Publication collection" and entity_type != "Publication dataset" %}
```
LavMatt, line 2: need to add charts and dashboards here. Wondering if there's a better way to do it though in case there are charts/dashboards added that are not public. Could we infer it via the `security_classification` which mitch has just added, where `Official` maps to the previous `audience: Published`

LavMatt: Just seen this also might change, https://github.com/ministryofjustice/find-moj-data/issues/1305

LavMatt (approved):

Issue #1287, opened 2025-01-27
Add Beta banner for the feedback form


teeceeas:
Add a Beta banner and include the link to the feedback form, so that we can improve feedback form interactions
https://design-system.service.gov.uk/components/phase-banner/

![Image](https://github.com/user-attachments/assets/375d3d90-8be5-4c93-9e05-cdfcbbf41c27)

PR #1321, opened 2025-01-30
design: add back the phase banner

MatMoore:
Previously removed in 713771d, but we want to make the feedback form more prominent, so we're adding it back in.

(update - this has been moved from below the main header to below the primary nav)

Screenshot 2025-01-30 at 13 54 04


joe-horton-moj:
Another alternative approach is to keep it below the service navigation but on the home page, we invert the colours so it can be contained in the masthead – like this example:

https://x-govuk.github.io/govuk-prototype-components/examples/masthead/

Screenshot 2025-01-30 at 13 28 35

joe-horton-moj:
Not to include the home link breadcrumb in the example of course – but just showing the inverted phase banner on the masthead.

MatMoore:
The inverted one looks good imo:

Screenshot 2025-01-30 at 13 54 04

Switching to this version of the component also reduces the amount of padding, which works better with the new design I think.

The other pages look like this:

Screenshot 2025-01-30 at 14 06 38

MatMoore:
Ticket: https://github.com/ministryofjustice/find-moj-data/issues/1287

joe-horton-moj:
Great, looking good

MatMoore:
As discussed, I've made a further change to reduce the spacing between breadcrumb & caption on the details page:

Screenshot 2025-02-03 at 14 55 04

LavMatt (approved):

PR #1401, opened 2025-02-18
refactor: use a template tag for pagination

MatMoore:
The previous partial hardcoded parts of the URL, which meant it could only be used on the search page. This change allows us to include pagination in other places, like the glossary.

The template tag takes two arguments: a `django.core.paginator.Page` object, and the name of a url pattern.

The page number is assumed to be in the path part of the url, rather than a query param, since this was the existing pattern. Any extra keywords are passed as keyword arguments to the url pattern, to support nested paths like `/glossary/{urn}/pagination/{page_number}`.


murdo-moj (approved):
It works!

PR #1402, opened 2025-02-19
feat: Enhance feedback and issue report forms with custom error messages and validation

murdo-moj:
* Add detail to error messages
* Move error box to the tope of the page
* Get the error boxes red bar working
* Issue form redirects to the entity with a confirmation banner
* Guard redirect against open redirect vulnerability

### DIFF
--- a/feedback/views.py
+++ b/feedback/views.py
@@ -1,5 +1,6 @@
 import logging
 
+from django.core.validators import MinValueValidator
 from django.http import HttpResponse
 from django.shortcuts import redirect, render
 
@@ -12,6 +13,9 @@ log = logging.getLogger(__name__)
 def feedback_form_view(request) -> HttpResponse:
     if request.method == "POST":
         form = FeedbackForm(request.POST)
+        form.fields['satisfaction_rating'].validators = [
+            MinValueValidator(1, message='Please enter a valid bid')
+        ]
         if form.is_valid():
             feedback = form.save()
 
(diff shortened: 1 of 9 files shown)


MatMoore:
Looks good, I just noticed that it errors when you open the form from the home page and submit successfully.

I think it might be due to it expecting an entity url that doesn't exist. Can we make the redirect default to the homepage?

feedback/views.py
```diff
@@ -69,8 +71,10 @@ def report_issue_view(request) -> HttpResponse:
                 issue=issue,
                 send_email_to_reporter=form.cleaned_data["send_email_to_reporter"],
             )
-
-            return redirect("feedback:thanks")
+            messages.add_message(
+                request, messages.SUCCESS, "Feedback submitted successfully"
+            )
+            return redirect(request.session.get("entity_url"))
```
MatMoore, line 77: ⚠️ there is a possibility here for an [open redirect vulnerability](https://learn.snyk.io/lesson/open-redirect/) if we don't validate this `entity_url` prior to redirecting the user, as this is derived from a request parameter.

To protect against that we can check that the URL is a relative one or starts with the service URL. I thought we had some code for this already but I can't find it now 🤔

murdo-moj:
I fixed the problem with the form submission from the homepage, I just haven't deployed it to dev yet

mitchdawson1982 (approved):
Nice

PR #1425, opened 2025-02-25
feat: Rearrange call to action for tables

murdo-moj:
* Tables and databases now have an "Access this data on the AP" section
* I've hardcoded in the database access page of the AP docs
* Fixed a bug where some relations of entities have no properties, which was causing an error

### DIFF
--- a/templates/details_table.html
+++ b/templates/details_table.html
@@ -5,25 +5,8 @@
 {% block extra_details %}
   <div class="govuk-grid-row">
     <div class="govuk-grid-column-full">
-      {% if entity.custom_properties.security_classification == "Official-Sensitive"%}
-      <div class="govuk-body govuk-!-padding-top-7 govuk-!-padding-bottom-7">
-        <h2 class="govuk-heading-m">Access this data</h2>
-        <p id="request_access" class="govuk-body"></p>
-          {% if entity.custom_properties.access_information.dc_access_requirements %}
-        {% if is_access_requirements_a_url %}
-          <a onclick="gtag('event', 'click_access_link', {'entity_name': '{{ entity.name }}'})" href="{{ entity.custom_properties.access_information.dc_access_requirements }}" class="govuk-link" rel="noreferrer noopener" target="_blank">Learn how to access this data in the Analytical Platform (opens in new tab)</a>
-        {% else %}
-          {{ entity.custom_properties.access_information.dc_access_requirements }}
-        {% endif %}
-        {% comment %}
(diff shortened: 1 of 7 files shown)


templates/details_table.html
```diff
@@ -5,6 +5,26 @@
 {% block extra_details %}
   <div class="govuk-grid-row">
     <div class="govuk-grid-column-full">
+      {% if entity.custom_properties.security_classification == "Official-Sensitive"%}
+      <div class="govuk-body govuk-!-padding-top-7 govuk-!-padding-bottom-7">
```
murdo-moj, line 9:
```suggestion
      <div class="govuk-body govuk-!-padding-top-7 govuk-!-padding-bottom-7">
```
I don't think this is actually body?

MatMoore: No doesn't look like it. I've stuck govuk-body on in a couple of divs where we are wrapping markdown generated content but that's not the case here.

templates/details_table.html
```diff
@@ -5,6 +5,26 @@
...
+        <h2 class="govuk-heading-m">Access this data</h2>
+        <p id="request_access" class="govuk-body"></p>
+          {% if entity.custom_properties.access_information.dc_access_requirements %}
```
MatMoore, line 12: the indenting of the L13 conditional is a bit off which makes this hard to read - can you make it indented one more level that L12?

templates/details_table.html
```diff
@@ -5,6 +5,26 @@
...
+        {% comment %}
+        {% elif entity.governance.data_custodians or entity.governance.data_owner.email %}
+        Contact the data custodian to request access.
+        {% endcomment %}
+          {% else %}
+        Not provided.
+          {% endif %}
+        </p>
+      </div>
+      {% endif %}
```
MatMoore, line 27: I don't think it's right for this logic to live in details_table, because we would have to duplicate it for databases or any other entity types that could contain official sensitive data.

We could simplify the template logic if we make it independent of entity type, and rely on `entity.custom_properties.security_classification` being set appropriately in the parsing code.

There is also the other variant of this handled in the `access_data.html` partial that is pulled in in for chart entities, which currently are assumed to always be public data. Can we combine these into a single partial, include it from the base template, and then have both cases as branches of the same conditional?

```
{% if entity.custom_properties.security_classification == "Official-Sensitive"%}
above case
{% else %}
... public case
{% end %}
```

MatMoore (dismissed):

PR #1436, opened 2025-02-27
Fix: Update entity_type in download button include for tables

mitchdawson1982:
- Fixes issue with the download of table metdata producing an empty csv file
- Adds pre-commit config update


murdo-moj (approved):
Nice!

PR #1440, opened 2025-02-28
Incorporate subject area tags in the front end

mitchdawson1982:


templates/glossary_term.html
```diff
@@ -21,7 +21,7 @@
 {% block content %}
 <div class="govuk-grid-row app-header-negative-padding">
   <div class="govuk-grid-column-full">
-    <span class="govuk-caption-l">Glossary term</span>
+    <span class="govuk-caption-l">Glossary</span>
```
MatMoore, line 24: Did we decide to keep this, or leave it as "Glossary term"?

mitchdawson1982: Joe wanted simply Glossary

PR #1451, opened 2025-03-05
Corrected royal crest

MalcolmVonMoJ:
@MatMoore 

This adds in the new Royal Crest SVG - ensure you have the backup PNG file in your assets, but I think you said that was automatically pulled in.


MatMoore (approved):
Thanks! The image was there, it's just we have a slightly different asset path.

murdo-moj (approved):

PR #1469, opened 2025-03-10
fmd-1414: Add Quality Assertions and logic to Parse into models:

hjribeiro-moj:
- Update grapqhql dataset query to get assertions from datahub
  - Create ColumnAssertions object and updated ColumnObject
  - Updated Parser to parse assertions and add it to columns

### DIFF
--- a/datahub_client/entities.py
+++ b/datahub_client/entities.py
@@ -115,6 +115,31 @@ Mappers = [
 ]
 
 
+class ColumnAssertionType(Enum):
+    COMPLETENESS = "Completeness"
+    CONSISTENCY = "Consistency"
+    NA = "N/A"
+
+
+class ColumnAssertionLevel(Enum):
+    RED = "red"
+    AMBER = "amber"
+    GREEN = "green"
+    NA = "N/A"
+
+
+class ColumnAssertionResult(Enum):
+    SUCCESS = "SUCCESS"
+    FAILURE = "FAILURE"
+    NA = "N/A"
+
+
+class ColumnAssertion(BaseModel):
+    type: ColumnAssertionType
+    level: ColumnAssertionLevel
+    # last_run: Optional[datetime]
+    result: ColumnAssertionResult
+
 class SecurityClassification(Enum):
     """Enumeration representing the security classification of the data.
     Attributes:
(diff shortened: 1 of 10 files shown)


murdo-moj:
Can you add a test case for `parse_assertions`?

murdo-moj:
I can see the DQ columns and they look good - the columns should only show for the one table we have assertions for. Currently they show for every table in the service, all of which are N/A

datahub_client/entities.py
```diff
@@ -115,6 +115,42 @@ class FindMoJdataEntityMapper:
...
+class ColumnAssertionResult(Enum):
+    SUCCESS = "SUCCESS"
+    FAILURE = "FAILURE"
+    NA = "N/A"
+
+
+class ColumnAssertion(BaseModel):
+    type: ColumnAssertionType
+    level: ColumnAssertionLevel
+    # last_run: Optional[datetime]
```
murdo-moj, line 142:
```suggestion
```

datahub_client/parsers.py
```diff
@@ -353,15 +388,28 @@ def parse_columns(self, entity: dict[str, Any]) -> list[Column]:
...
         for field in schema_metadata.get("fields", ()):
             foreign_keys_for_field = foreign_keys[field["fieldPath"]]
 
-            # Work out if the field is primary.
-            # This is an oversimplification: in the case of a composite
-            # primary key, we report that each component field is primary.
             is_primary_key = field["fieldPath"] in primary_keys
             field_path = field["fieldPath"]
             display_name = field_path.split(".")[-1]
+            column_assertions = all_column_assertions.get(display_name, {})
```
murdo-moj, line 398: what do you think about turning this into a function/method or including the logic in `parse_assertions`?

templates/details_table.html
```diff
@@ -61,5 +71,39 @@ <h2 class="govuk-heading-m">Lineage</h2>
           </a>
         </div>
       {% endif %}
+      
     </div>
+  </div>
+
+  <style>
+    /* GOV.UK style status colors */
```
murdo-moj, line 79: Can you move the styling to the custom CSS file we have?

datahub_client/parsers.py
hjribeiro-moj: the problem is that we have 2 separate blocks: assertions and schema (columns). For us to do it all-in-one, it would be per column and we'd be iterating possibly N! times, where N is the number of columns. I have refactored the code you highlighted into a function for readability.

templates/details_table.html
hjribeiro-moj: done

murdo-moj (approved):
💯

PR #1479, opened 2025-03-14
Fix failing integration tests

mitchdawson1982:
Adds a sleep method to the Page object to bypass issues with Staleobject exceptions. I have attempted various combinations of expected conditions without consistent success so far.


hjribeiro-moj (approved):
LGTM

PR #1494, opened 2025-03-20
fmd-1414-Fix RAG fonts on quality metrics

hjribeiro-moj:
Created a mapping between (poor, acceptable, good) -> (red, yellow, green) to import gov-uk tags.

Deleted the custom css that was created for that purpose.


mitchdawson1982 (approved):
Nice

PR #1499, opened 2025-03-21
Control reporting of entities doesn't exist errors to Sentry

mitchdawson1982:
Adds
- `before_send` helper function to the sentry sdk init `before_send` argument, to filter instances of `Http404` exceptions containing `does not exist`

This approach can be extended to include additional exception references such as invalid/missing type result.

Removes
- Redundant custom 404 exception handler view. This has been handled by custom exception middleware for sometime.


murdo-moj (approved):
Works great :)

PR #1960, opened 2025-09-29
Analytics Changes

mkjmckinlay:
Added GTM so we can track tag changes etc.

Added checks to the base template so if the user is staff we dont display the analytics codes.


core/settings.py
```diff
@@ -213,10 +213,11 @@
 CACHES = generate_cache_configuration()
 
 ANALYTICS_ID: str = os.environ.get("ANALYTICS_ID", "")
+GTM_ID: str = os.environ.get("GTM_ID", "")
 ENABLE_ANALYTICS: bool = (
     os.environ.get("ENABLE_ANALYTICS") in TRUTHY_VALUES
 ) and ANALYTICS_ID != ""
-
+STAFF = str("STAFF", ).split(",")
```
mitchdawson1982, line 220: Where is `STAFF` coming from?

core/settings.py
```diff
@@ -213,10 +213,11 @@
 CACHES = generate_cache_configuration()
 
 ANALYTICS_ID: str = os.environ.get("ANALYTICS_ID", "")
+GOOGLE_TAG_MANAGER_ID: str = os.environ.get("GOOGLE_TAG_MANAGER_ID", "")
 ENABLE_ANALYTICS: bool = (
     os.environ.get("ENABLE_ANALYTICS") in TRUTHY_VALUES
 ) and ANALYTICS_ID != ""
-
+STAFF = str(os.environ.get("STAFF"), ).split(",")
```
mitchdawson1982, line 220:
```suggestion
STAFF = str(os.environ.get("STAFF", "")).split(",")
```

mitchdawson1982 (dismissed):
LGTM

PR #2088, opened 2025-12-02
Ga fixes

mkjmckinlay:
Please add the STAFF list and Google Analytics config values  to 1Password.


templates/base/base.html
```diff
@@ -30,21 +29,20 @@
 </head>
 
 <body class="govuk-template__body app-body-class" data-test="My value" data-other="report:details">
-{% if request.user not in STAFF %}
+  {% if request.user|lower not in STAFF %}
```
mitchdawson1982:
```suggestion
  {% if request.user.email|lower not in STAFF %}
```
If the str representation of user ever changed this could potentially no longer match. I suggest it is better to be specific to the required field.

PR #2382, opened 2026-05-27
Found in

mkjmckinlay:


home/service/search.py
```diff
@@ -184,6 +184,25 @@ def _get_context(self) -> dict[str, Any]:
...
+                "fieldDescriptions",
+                "id",
+                "urn",
+                "tags",
+                "dbt_file_path",
+                "dbt_unique_id",
+                "name",
+                "domain",
+                "database",
+                "tags",
```
nathan-justice: I think 'tags' is already mentioned once - second could be removed?

nathan-justice (changes requested):

home/service/search.py
mkjmckinlay: Good catch!

nathan-justice (approved):

PR #2398, opened 2026-06-08
Add `aria-current="page"` to active pagination item for accessibility…

mkjmckinlay:
… compliance


templates/partial/govuk/pagination.html
```diff
@@ -14,7 +14,7 @@
                 <li class="govuk-pagination__item govuk-pagination__item--ellipses">&ctdot;</li>
             {% else %}
                 {% if page_obj.number == p.number %}
-                    <li class="govuk-pagination__item govuk-pagination__item--current">
+                    <li class="govuk-pagination__item govuk-pagination__item--current" aria-current="page">
```
nathan-justice: Could be wrong, but in another file in FMD the `aria-current="page"` is in the `<a>` anchor line - not sure if that should be the case here?
File **navigation.py**:
`<a class="moj-primary-navigation__link" href="{% url 'home:search' %}?new=True" {% if request.path == search_url %}aria-current="page"{% endif %}>`
I don't know how to test with a screen-reader... 0-0

nathan-justice (changes requested):

nathan-justice (approved):

